PR #4, opened 2018-09-19
Backward incompatible v2

stalker314314:
Let's make things official, moving to master:) Please do skim over the code, it would mean a lot!;)

### DIFF
--- a/appinfo/database.xml
+++ b/appinfo/database.xml
@@ -44,16 +44,16 @@
 			</field>
 			<!-- foreign key: users.uid -->
 			<field>
-				<name>user_id</name>
+				<name>user</name>
 				<type>text</type>
 				<notnull>true</notnull>
 				<length>64</length>
 			</field>
 			<index>
-				<name>persons_user_id_idx</name>
+				<name>persons_user_idx</name>
 				<unique>false</unique>
 				<field>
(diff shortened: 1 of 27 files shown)


matiasdelellis:
Hi @stalker314314 ,
At this point, I have no comments about the code. It seems to be correct. - In the task of creating and stabilizing the groups I will trust you -. But I would like to know how you will continue. :wink:

stalker314314:
> About creating and stabilizing the clustera I will trust you -. 

Huh, great, as I am not trusting myself either:) I have no idea how this will all turn out, I just hope for the best🤪

> But I would like to know how you will continue. 😉

Not sure if this is rethorical question or not, but here is high-level plan:
* Add support for hooks
* Either duplicate frontend or change it to work with new DB (will have to figure out)
* Move all TODO from code to GitHub issues and crunch them, one by one
* Tests and more tests
* Some CI pipeline
* Make it super easy for people to install this (and pdlib and whats not)
* Documentation
* Ask for feedback
* Release to NC store

In any case, I need to be **as least destructive to current (working) code as possible**! At least until we can say with high certainty that this thing actually works.

I guess I can merge pull request on my own? Will do it tomorrow 😄

PR #6, opened 2018-09-26
Elegantly excuse some mistakes with shared files.

matiasdelellis:
The path formed by shared files is wrong.

With this patch, without fixing the main problem, at least continue the analysis just showing a message of the error (and try to read it again in the next analysis).
However, keep in mind, that the patch is still valid, for example when the files for some reason are corrupt.

### DIFF
--- a/lib/BackgroundJob/Tasks/ImageProcessingTask.php
+++ b/lib/BackgroundJob/Tasks/ImageProcessingTask.php
@@ -78,9 +78,15 @@ class ImageProcessingTask extends FaceRecognitionBackgroundTask {
 			$userRoot = $userFolder->getParent();
 			$file = $userRoot->getById($image->file);
 			// todo: handle case where file do not exists anymore
+			// todo: this concat is wrong with shared files.
 			$imagePath = $dataDir . $file[0]->getPath();
 			$this->logInfo('Processing image ' . $imagePath);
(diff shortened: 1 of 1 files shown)


lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -78,10 +78,16 @@ public function do(FaceRecognitionContext $context) {
 			$userRoot = $userFolder->getParent();
 			$file = $userRoot->getById($image->file);
 			// todo: handle case where file do not exists anymore
+			// todo: this concat is wrong with shared files.
 			$imagePath = $dataDir . $file[0]->getPath();
 			$this->logInfo('Processing image ' . $imagePath);
 			list($tempfilePath, $ratio) = $this->prepareImage($imagePath);
 
+			if ($tempfilePath == null || $ratio <= 0.0) {
+				$this->logInfo('There was some error with the file, continue with the next one');
```
stalker314314: in some future version, we should update Image to add it is a) processed, b) stack trace to it (so it doesn't always go over same file). But, yes, it is OK to unblock both you and me now!

stalker314314:
I guess it is OK for me to do merge, since I reviewed it?:) Let me know if you don't want me to do this in future:)

matiasdelellis:
Of course Branko, :+1: 
The best thing is that we check the pull request mutually.. :wink:

PR #7, opened 2018-09-29
Hooks v2 implementation

stalker314314:
This change implements add/modify/delete file hooks, as per functional spec (or similar to it:) I am not touching existing hooks.

When file is added, it is just added to `Image` DB table. If file is changed, it is reset in `Image` table, and all persons that had faces from that image are set is `is_valid=false` (so, that we signal person is not up-to-date). When file is deleted, we do the same, but remove it from `Image` table.

Also experimenting how we get `$owner`. Instead of:
```
$absPath = ltrim($file->getPath(), '/');
$owner = explode('/', $absPath)[0];
```
I am trying with:
```
$owner = $node->getOwner()->getUid();
```

Maybe it's the same, not sure.

Also blocked any callback for modifications for other storages (shared, external...) as they don't have reliable callbacks (but have to dig more on this).

Some small bug fixes in this, too

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -43,34 +43,37 @@ class Application extends App {
 
 		$container = $this->getContainer();
 
-		$this->connectWatcher($container);
+		$this->connectWatcher();
 		$this->connectSearch($container);
 		$this->registerAdminPage();
 	}
 
-	/**
-	 * @param IAppContainer $container
-	 */
-	private function connectWatcher(IAppContainer $container) {
+	private function connectWatcher() {
 		/** @var IRootFolder $root */
-		$root = $container->query(IRootFolder::class);
-		$root->listen('\OC\Files', 'postWrite', function (Node $node) use ($container) {
+		$root = $this->getContainer()->query(IRootFolder::class);
+		$root->listen('\OC\Files', 'postWrite', function (Node $node) {
 			/** @var Watcher $watcher */
-			$watcher = $container->query(Watcher::class);
+			$watcher = \OC::$server->query(Watcher::class);
 			$watcher->postWrite($node);
+			$watcher->postWritev2($node);
 		});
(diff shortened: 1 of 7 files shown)


matiasdelellis:
Hi @stalker314314 

> This change implements add/modify/delete file hooks, as per functional spec (or similar to it:) I am not touching existing hooks.

Well, We could already break comparibility because in any time will have two applications in the same repository.  :sweat_smile: 

> When file is added, it is just added to `Image` DB table. If file is changed, it is reset in `Image` table, and all persons that had faces from that image are set is `is_valid=false` (so, that we signal person is not up-to-date). When file is deleted, we do the same, but remove it from `Image` table.

It is not a priority, but to investigate, we must find out exactly when the HOOK is generated. Maybe it is generated by several events that should not invalidate the image

> Also experimenting how we get `$owner`. Instead of:
> 
> ```
> $absPath = ltrim($file->getPath(), '/');
> $owner = explode('/', $absPath)[0];
> ```
> 
> I am trying with:
> 
> ```
> $owner = $node->getOwner()->getUid();
> ```
> 
> Maybe it's the same, not sure.

Ohh .. Beautiful .. :smiley: 
I copied that from some application, and always seemed prone to errors.

> Also blocked any callback for modifications for other storages (shared, external...) as they don't have reliable callbacks (but have to dig more on this).

Great. Maybe later we can manage them, but now it is not a priority!

> Some small bug fixes in this, too

 :wink: 

Well,
Now see the code.

Thank for all!

matiasdelellis:
> It is not a priority, but to investigate, we must find out exactly when the HOOK is generated. Maybe it is generated by several events that should not invalidate the image

I mean to "change" hook. :wink:

matiasdelellis:
About the code, I do not have comments,
Tomorrow I do a few tests on the server, but if you are confident merge it. :wink:

PR #30, opened 2018-10-06
Determine mime types in central place, fixes #27

### LINKED ISSUE #27: Choose file type we are supporting
We have in code sprinkled support for image/jpeg and/or image/png. Some questions:
* should we decide file type based on mime type from Nextcloud? Probably yes
* what mimetypes Nextcloud can handle?
* what types dlib handles?
* what happens if we add support for more filetypes? Should we kick in new `AddMissingImagesTask` run? Reset FULL_IMAGE_SCAN_DONE_KEY in migration?

stalker314314:


lib/BackgroundJob/Tasks/AddMissingImagesTask.php
```diff
@@ -161,8 +162,7 @@ private function getPicturesFromFolder(Folder $folder, $results = array()) {
 			if ($node instanceof Folder and !$node->nodeExists('.nomedia')) {
 				$results = $this->getPicturesFromFolder($node, $results);
 			} else if ($node instanceof File) {
-				// todo: which are filetypes we can work with
-				if ($node->getMimeType() === 'image/jpeg') {
+				if (Requirements::isImageTypeSupported($node->getMimeType())) {
```
matiasdelellis: Do you know how to use dependency injection in these cases?.
Honestly did not try ..

lib/Helper/Requirements.php
```diff
@@ -55,4 +55,29 @@ public function getLandmarksModel ()
...
 	}
 
+	/**
+	 * Determines if FaceRecognition can work with a givem image type. This is determined as
+	 * intersection of types that are supported in Nextcloud and types that are supported in DLIB.
+	 *
+	 * Dlib support can be found here:
+	 * https://github.com/davisking/dlib/blob/9b82f4b0f65a2152b4a4243c15709e5cb83f7044/dlib/image_loader/load_image.h#L21
+	 *
+	 * Note that Dlib supports these if it is compiled with them only! (with libjpeg, libpng...)
```
matiasdelellis: It's ok..
If then it returns some exception it is already "elegantly" managed.

lib/BackgroundJob/Tasks/AddMissingImagesTask.php
stalker314314: Dependency injection when using static methods? Not possible:D (static methods are antipattern for DI).
If we ever plan to mock "isImageTypeSupported", it will be impossible. However, I don't see us using different implementation of this method. If we ever do, we should just make it non-static, I guess

PR #31, opened 2018-10-17
Image processing task robustness

### LINKED ISSUE #17: Exception handling in ImageProcessingTask
Number of things that can go wrong in this logic is just too damn high.

https://knowyourmeme.com/memes/the-rent-is-too-damn-high-jimmy-mcmillan

### LINKED ISSUE #19: Make sure temp images are deleted in finally block
In `ImageProcessingTask`

### LINKED ISSUE #20: CropFace is loading same image two times for no reason (perf issue)
In `ImageProcessingTask`

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/64296801-cropface-is-loading-same-image-two-times-for-no-reason-perf-issue?utm_campaign=plugin&utm_content=tracker%2F74944432&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F74944432&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #21: resizeImage in ImageProcessingTask should throw exception on error

stalker314314:
Handled exception are now written to DB, if image cannot be processed
and error is written to DB. Unhandled exception continue to block whole job.

This should fix #17, fix #19, fix #21

### DIFF
--- a/lib/BackgroundJob/Tasks/ImageProcessingTask.php
+++ b/lib/BackgroundJob/Tasks/ImageProcessingTask.php
@@ -31,8 +31,61 @@ use OCP\IUser;
 
 use OCA\FaceRecognition\BackgroundJob\FaceRecognitionBackgroundTask;
 use OCA\FaceRecognition\BackgroundJob\FaceRecognitionContext;
+use OCA\FaceRecognition\Db\FaceNew;
 use OCA\FaceRecognition\Db\Image;
 use OCA\FaceRecognition\Db\ImageMapper;
+
+/**
+ * Plain old PHP object holding all information
+ * that are needed to process all faces from one image
+ */
+class ImageProcessingContext {
+	/** @var string Path to the image being processed */
+	private $imagePath;
+
+	/** @var string Path to temporary, resized image */
+	private $tempPath;
+
+	/** @var float Ratio of resized image, when scaling it */
+	private $ratio;
+
+	/** @var array<FaceNew> All found faces in image */
+	private $faces;
+
+	public function __construct(string $imagePath, string $tempPath, float $ratio) {
+		$this->imagePath = $imagePath;
+		$this->tempPath = $tempPath;
+		$this->ratio = $ratio;
+	}
+
+	public function getImagePath(): string {
+		return $this->imagePath;
+	}
+
+	public function getTempPath(): string {
+		return $this->tempPath;
+	}
+
+	public function getRatio(): float {
+		return $this->ratio;
+	}
+
+	/**
+	 * Gets all faces
+	 *
+	 * @return FaceNew[] Array of faces
+	 */
+	public function getFaces(): array {
+		return $this->faces;
+	}
+
+	/**
+	 * @param array<FaceNew> $faces Array of faces to set
+	 */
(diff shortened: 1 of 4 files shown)


lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -70,66 +123,88 @@ public function do(FaceRecognitionContext $context) {
 		$fld = new \FaceLandmarkDetection("shape_predictor_5_face_landmarks.dat");
 		$fr = new \FaceRecognition("dlib_face_recognition_resnet_model_v1.dat");
 
-		// number of things that can go wrong below is too damn high:) be more defensive
 		foreach($images as $image) {
+			yield;
```
stalker314314: moved yield to start. There was some good reason (not to be after finally), but I forgot (but, there was a reason!:)

lib/Db/ImageMapper.php
```diff
@@ -108,16 +108,29 @@ public function findImagesWithoutFaces(IUser $user = null) {
...
+	 */
+	public function imageProcessed(Image $image, array $faces, int $duration, \Exception $e = null) {
 		$this->db->beginTransaction();
-		$currentDateTime = new \DateTime();
 		try {
 			// Update image itself
 			//
+			$error = null;
+			if ($e != null) {
+				$error = substr($e->getMessage() . "\n" . $e->getTraceAsString(), 0, 1024);
```
stalker314314: is there a better way to format exception to get both message and trace?

matiasdelellis: I do not know.. Maybe these..

https://github.com/nextcloud/3rdparty/blob/b995ca8b8c7f69a180ad6fb49989ad3e35b1367e/pear/pear_exception/PEAR/Exception.php#L443
https://github.com/nextcloud/lookup-server/blob/10e5382215e55a9d419b51db265fed0fa11f1f6d/server/vendor/slim/slim/Slim/Handlers/Error.php#L163

However, saving the backtrace in db IMHO seems unnecessary. I see better to keep a simple integer that represents an known enumerated type, so that it is simple to consider, when you want to analyze the file again.

I mean, the backtrace seems important to me for the nextcloud log.. For that the administrator can take actions to fix it, but not for the general operation of the application..

stalker314314: @matiasdelellis  I was also thinking in direction to have predictable hash here, for example `md5(some fixed part of stack trace)`🤪 (because you don't care about actual exception and stack trace, only about doing aggregates + it is easier to send it outside, as it is anonymous).

In any way, I think you are right - I was using database as a logging facility! This should end up in log and not in DB. If you are OK that `error` column stays as `text`, I would write here just `$e->getMessage()`. If you think **nothing** should exists, I can change `error` column to `bool`? I prefer keeping it as `text`, so we can have at least *something* in here (I do not think it should be exposed to user from frontend anyway).

That being said, I will add new commit that moves this to lightweight variant to log only message. Push if you are OK with it, comment if you are not:)

PR #33, opened 2018-10-22
Prevent running from ajax, closes #13

### LINKED ISSUE #13: Don't allow background jobs from AJAX cron jobs
> (TODO in BackgroundService.php)

We should disallow AJAX cron jobs to be able to run background jobs. This defies whole purpose of *background* job (notice what is highlighted:smile:) But, somewhere in admin settings, there should be warning telling admin that background server (and therefore whole app) is **not** working.

stalker314314:


lib/BackgroundJob/Tasks/CheckCronTask.php
```diff
@@ -0,0 +1,57 @@
...
+		return "Check that service is started from either cron or from command";
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	public function do(FaceRecognitionContext $context) {
+		$this->setContext($context);
+
+		$isCommand = $context->isRunningThroughCommand();
```
matiasdelellis: This is not always true?
You set that in https://github.com/matiasdelellis/facerecognition/pull/33/files#diff-3eec1bf8e625d99e338e741f9c524b9cR83

stalker314314: So, I am keeping default value as `false` since (once we have background service), it will not be possible to set this in background service (background service class will be `BackgroundService.php`. So, this means that, in command, we have opportunity to set this to true and I am doing that in [line 83 here](https://github.com/matiasdelellis/facerecognition/pull/33/files#diff-3eec1bf8e625d99e338e741f9c524b9cR83). This is how we differentiate between command and service (this is my thinking).

matiasdelellis: Ok.. ;)

PR #36, opened 2018-10-24
Mimetype change is not problem, closes #28

### LINKED ISSUE #28: Handle case when in postWrite file changes mimeType
In `postWritev2`, if file was JPEG, but was overwritten with, let's say .txt, we have problems. Also handle other way around (.txt->.jpg)

stalker314314:
So, it turns out, I was wrong in my assumptions. Mimetype is not
determined in some automagical fashion, but simply as extension from
filename. So, that means - there was no problem and I just removed TODO
items in this commit:)

Simple test to deduce is:
* Create TXT file with some text in it and rename it as foo.jpg
* Upload to nextcloud
* Detect from DB (oc_mimetypes) that mimetype is image/jpeg (this means
* NC is not smart to peek into content and deduce mimetype)

### DIFF
--- a/lib/Watcher.php
+++ b/lib/Watcher.php
@@ -142,7 +142,6 @@ class Watcher {
 			return;
 		}
 
-		// todo: what if file was image and now its mimetype is changed?:) we should remove it
 		// todo: honor .nomedia here
 		if (!Requirements::isImageTypeSupported($node->getMimeType())) {
 			return;
@@ -213,7 +212,6 @@ class Watcher {
 			return;
 		}
 
-		// todo: what if file was image and now its mimetype is changed?:) we should remove it
 		if (!Requirements::isImageTypeSupported($node->getMimeType())) {
 			return;
 		}


matiasdelellis:
Wow, You are with all the batteries. I promise that soon I start with B! :smile: 

Thinking of the 'change' HOOK, is ok.. So merge that..

Now I have doubts about the HOOK of new files.. Since if you upload a text file with jpg extension, this is already considered image/jpeg.. However, I am inclined to take all the images that nexcloud considers images, and in any case exit elegantly when there is an error of that style.

stalker314314:
When we notice TXT file with .jpg extension, we will try resizing it, searching for faces, but it will fail. We will add error in `error` column in `oc_face_recognition_images` and that's it. That "image" is ignored forever, does not impact performance. But it will be checked by us, because we rely on NC to tell us is it image or not. Is this what you are asking/wanting?

And vica versa, if we upload image renamed as foo.txt, it will be ignored by us. I guess this is OK, I don't think we should be in the business of detecting mime types:)

matiasdelellis:
Of course. We trust in nextcloud to get the mimetypes.. and try to open once, and act accordingly. :wink: 
The comment comes in that I honestly thought that nextcloud was smarter. hahah..

Issue #37, opened 2018-10-24
Add task to remove images that do not exist physically, but are in DB


stalker314314:
This task will be needed only if user adds file `.nomedia` in some folder and all images in that folder and in child folders should be removed from our DB!

But, instead of doing this in hook (it could be very, very expensive), my proposal is to do it in new task - something like `RemoveMissingImages`. This task should take all images from DB, check that they actually exist (and no parent above them have `.nomedia`) and remove them from DB if this is not true (using same logic as delete hook in `Watcher.php`). This task should be guarded with some config value (similar to `AddMissingImagesTask::FULL_IMAGE_SCAN_DONE_KEY`), where default value should be `false`, meaning this task is not run, unless `postWritev2` detects that `.nomedia` was added in some folder and set this to `true`, so next background job will execute this.

PR #38, opened 2018-10-25
Handle (partially) .nomedia cases, fixes #29

### LINKED ISSUE #29: Handle .nomedia in Watcher
@matiasdelellis added support for .nomedia. It looks like a good feature, but we need to honor it in `Watcher.php`. Seems that we can accidentally add images from .nomedia folder (or child of it).

stalker314314:
Fixes cases when image is modified inside some directory which has
`.nomedia` in it, or some of its ancestor dirs have .nomedia.

Also fixes case when `.nomedia` file (itself!) is deleted, as this means all
images from this directory (and child directories) are now eligable for face
detection. However, similar case (when `.nomedia` file is added to some
directory) is not handled and is just referenced as issue #37. This one
requires new task to be added.


matiasdelellis:
Oh great !.
Sorry for the delay ..

PR #39, opened 2018-11-02
Central place for loading DNN models, fixes #15

### LINKED ISSUE #15: Central place for loading DNN models
In `ImageProcessingTask` class, there are loading of models, which should be better if they are in `Requirements` class, or somewhere else, central to where we can find it (with more complex lookup logic to find models). This task also depends on how we load and keep models (apt-get/manually, or automaticall in future)

stalker314314:
This change adds support to download models in Makefile (and adds it as
default target in make).

Requirement task is renamed to check not only does pdlib is loaded, but
also if model files are present.

Requirement class is changed, so modelId is required argument for it to
operate. It encapsulate logic to retrieve paths to model files based on
model ID (which currently does not make any sense:) as we have only one
model), but yes - Requirement now hides this from developer.

### DIFF
--- a/models/README.md
+++ b/models/README.md
@@ -1,0 +1,8 @@
+This directory holds all model files which are required by facerecognition app to work properly.
+
+Each set of files from a model is in separate directory, where directory name is identified of the model.
+
+You should at least have one directory named '1/' in here. If this directory is not present, or it is empty
+(there are no models in it), try executing `make download_models` from root directory of the application.
+
+If this doesn't help, file an issue at: https://github.com/matiasdelellis/facerecognition/issues
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ opt
 js/handlebars.js
 js/lozad.js
 vendor/
+models/1
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,26 @@ source_dir=$(build_dir)/source
 package_name=$(app_name)
 cert_dir=$(HOME)/.nextcloud/certificates
 
-deps:
+default: deps
+
+models/1/mmod_human_face_detector.dat:
+	mkdir -p models/1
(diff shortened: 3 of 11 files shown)


matiasdelellis:
Perhaps the license must be distributed, or at least referenced in the readme... but about code I do not have any comments... Greats!. :+1:

stalker314314:
You mean, in our README.md? I can point to [this link](https://github.com/davisking/dlib-models) and quote Davis saying:
>  As far as I am concerned, anyone can do whatever they want with these model files as I've released them into the public domain. 

Only _problematic_ one is shape_predictor_68_face_landmark (not problematic, but requires attribution), but we don't use/download/need that one

Just confirm that you want me to add to README.md licences? (I am not a lawyer:)

models/README.md
```diff
@@ -0,0 +1,8 @@
+This directory holds all model files which are required by facerecognition app to work properly.
+
```
matiasdelellis: Hi @stalker314314 
A link here as a reference may be enough,,

Ideally in models/1 along with the files, but it can be generic. Maybe something like that here:

 * The models used for facecerecognition are extracted from https://github.com/davisking/dlib-models
 * General license: https://github.com/davisking/dlib-models/blob/master/LICENSE
 * See details on: https://github.com/davisking/dlib-models/blob/master/README.md

stalker314314: Yes, reference outside of main README.md seems better. However, I opted to create separate, LICENCE file in `models/` dir, not use eixsting `models/README.md`. If you agree, merge pull request please:) If not, I can move to `models/README.md`.

matiasdelellis:
Cool.. :+1: 
I'm not a lawyer either, but if we're going to distribute it, we should attribute the licenses ..

PR #40, opened 2018-11-10
Relying to ITempManager from NC, fixes #18

### LINKED ISSUE #18: Reimplement choosing temp folder logic
Currently, we use machine's temp folder to store resized versions of images. This is not secure, We need to leverage NC capabilities to have temp folders. @matiasdelellis had comment and also idea here:

https://github.com/matiasdelellis/facerecognition/pull/4#commitcomment-30244624

stalker314314:
This doesn't fix original issue stated by Matias (that this is not secure) since I learned during testing that ITempManager also uses global temp directory (`/tmp`), but at least we are relying more on Nextcloud with this commit `¯\_(ツ)_/¯`.

TBH, I am not sure what _secure_ means here, as - even if we are using some Nextcloud temp directory, whoever has access to machine will be able to read those files. Moving stuff to `/tmp` doesn't change anything. I am not expert, but I think we will have to live with `/tmp` :/

### DIFF
--- a/lib/BackgroundJob/Tasks/ImageProcessingTask.php
+++ b/lib/BackgroundJob/Tasks/ImageProcessingTask.php
@@ -28,6 +28,7 @@ use OC_Image;
 use OCP\Files\File;
 use OCP\Files\Folder;
 use OCP\IConfig;
+use OCP\ITempManager;
 use OCP\IUser;
 
 use OCA\FaceRecognition\BackgroundJob\FaceRecognitionBackgroundTask;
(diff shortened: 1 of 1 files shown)


matiasdelellis:
Great!.
About security, the problem is if someone gains access to your server. Since /tmp is public, it has permissions, anyone can see and use it.. Highly unlikely and can also be corrected with permissions. ;)

My concern was that in shared environments can be read-only, and it does not work directly. Probably this is only in some old virtualization environments.. So, we can trust NC core... :wink:

PR #45, opened 2018-11-12
Migrate admin panel to new standard. Fix #42

### LINKED ISSUE #42: RegisterAdmin call is deprecated, create proper settings
Current approach is to create admin page with:
```
\OCP\App::registerAdmin
```
and this is deprecated:
```
@deprecated 14.0.0 Use settings section in appinfo.xml to register admin sections
```

This is blocking ./occ app:check to work in continuous integration

I guess we need to rewrite admin page to be in new (NC>=v10) format.

matiasdelellis:
Implement the new panel, and leave the previous one for Nc <14

It remains so basic, and only shows the installation of the models and pdlib..
As a regression, but it's not a priority, it does not show progress, and I'm going to ask you for help to implement the new process controller.


lib/AppInfo/Application.php
```diff
@@ -80,7 +80,8 @@ private function connectSearch() {
 	 * Register admin settings
 	 */
 	public function registerAdminPage() {
-		\OCP\App::registerAdmin($this->getContainer()->getAppName(), 'templates/settings');
+		if (\OCP\Util::getVersion() < 14)
```
stalker314314: I don't think you need this:
1) `.\occ app:check` will fail when it sees `\OCP\App::registerAdmin` (as it is deprecated), and
2) Even v11 (which is our minimum version) supports <settings><admin>, as stated [here](https://docs.nextcloud.com/server/11/developer_manual/app/info.html)

Can we just completely remove `registerAdminPage` method?

templates/admin.php
```diff
@@ -0,0 +1,23 @@
...
+<form id="facerecognition">
+	<div class="section">
+		<h2>
+			<?php p($l->t('Face recognition'));?>
+			<a target="_blank" rel="noreferrer noopener" class="icon-info" title="<?php p($l->t('Open Documentation'));?>" href="https://github.com/matiasdelellis/facerecognition/wiki"></a>
+		</h2>
+		<h3>Configuration information</h3>
+		<p><strong>Requirements: </strong><?php p($_['msg']); ?><span class="status success<?php if(!$_['requirements']):?> error<?php endif;?>"></span></p>
+		<h3>Models</h3>
+		<p><strong>Recognition Model: </strong><?php p($_['recognition-model']); ?></p>
```
stalker314314: Consider (not required, just think about it):
Maybe we should not give user full path to models and what each model does. Maybe it is enough to just tell them something like "Using model: "1 (default)" and whether model files are present or not (as requirements). This is because it is unnecesarry information for user (why they need to know location of model files?) and maybe model "2" will not be having same set of files, maybe model 2 have "face_recognition", "baby_recognition", "nose_detector"... Again, whatever is easier/makes sense to you.

stalker314314:
I approved, but I will wait for couple of days, to hear back from you, if you plan to make any additional changes:) If I don't see from you soonish, I'll just merge:D

lib/AppInfo/Application.php
matiasdelellis: > Can we just completely remove registerAdminPage method?

Of course.. Just leave it, because in all your commits you are working in parallel, leaving the previous code. 
For my part, I want to leave behind all vestige of the python tool ..

templates/admin.php
matiasdelellis: You are right..
Considering that we now distribute the models in the application, it does not make sense to say what these models are.

lib/Settings/Admin.php
```diff
@@ -40,47 +40,35 @@ public function getSection() {
...
 
-		$model = intval($this->config->getAppValue('facerecognition', 'model', AddDefaultFaceModel::DEFAULT_FACE_MODEL_ID));
+		$modelVersion = intval($this->config->getAppValue('facerecognition', 'model', AddDefaultFaceModel::DEFAULT_FACE_MODEL_ID));
 
-		$req = new Requirements($this->appManager, $model);
+		$req = new Requirements($this->appManager, $modelVersion);
 
-		if (!$req->pdlibLoaded()) {
-			$requirements = FALSE;
-			$msg += 'PDLib is not loaded. Configure it';
```
matiasdelellis: WTF? This is not python!. haha.. :man_facepalming:

matiasdelellis:
Done... Try it and if it convinces you, merge .. :smile:

stalker314314:
I think we couldn't enable app:check just yet:) Take a look at this:
https://travis-ci.org/matiasdelellis/facerecognition/jobs/463229113
I will disable it later today, and maybe I will try cleaning it altogether, but not sure if we can get 100%. I will see.

matiasdelellis:
Hi,

> I think we couldn't enable app:check just yet:) Take a look at this:
> https://travis-ci.org/matiasdelellis/facerecognition/jobs/463229113
> I will disable it later today, and maybe I will try cleaning it altogether, but not sure if we can get 100%. I will see.

I'm not 100% sure, but we can make report the errors and continue analyzing in this way.

 * https://github.com/nextcloud/firstrunwizard/blob/master/.drone.yml#L34
 * https://support.cloudbees.com/hc/en-us/articles/218517228-How-to-Ignore-Failures-in-a-Shell-Step-

matiasdelellis:
> Hi,
> 
> > I think we couldn't enable app:check just yet:) Take a look at this:
> > https://travis-ci.org/matiasdelellis/facerecognition/jobs/463229113
> > I will disable it later today, and maybe I will try cleaning it altogether, but not sure if we can get 100%. I will see.
> 
> I'm not 100% sure, but we can make report the errors and continue analyzing in this way.
> 
>     * https://github.com/nextcloud/firstrunwizard/blob/master/.drone.yml#L34
> 
>     * https://support.cloudbees.com/hc/en-us/articles/218517228-How-to-Ignore-Failures-in-a-Shell-Step-

Hi @stalker314314 
Momentarily we can do that.. Report errors, but do not stop all the subsequent analysis ..:wink:

https://travis-ci.org/matiasdelellis/facerecognition/jobs/463381887

PR #46, opened 2018-11-13
Move search to new db scheme

matiasdelellis:
Returning to the origins.. haha..
This was the only way to access photos of a person at the beginning.. :grin: 

..and it is a first way to evaluate the groups of your new logic. :wink: 

Note that it depends on NC14 -findEntities()- , and and php7 -Return Type Declarations- ... :man_shrugging:

### DIFF
--- a/lib/Db/ImageMapper.php
+++ b/lib/Db/ImageMapper.php
@@ -108,6 +108,26 @@ class ImageMapper extends Mapper {
 		return $images;
 	}
 
+
+	public function findImagesFromPerson(string $userId, string $name, $model): array {
+		$qb = $this->db->getQueryBuilder();
+		$query = $qb
+			->select('i.id', 'i.file')
+			->from('face_recognition_images', 'i')
+			->innerJoin('i', 'face_recognition_faces', 'f', $qb->expr()->eq('f.image', 'i.id'))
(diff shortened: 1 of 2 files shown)


lib/Db/ImageMapper.php
```diff
@@ -108,6 +108,26 @@ public function findImagesWithoutFaces(IUser $user = null) {
 		return $images;
 	}
 
+
+	public function findImagesFromPerson(string $userId, string $name, $model): array {
```
stalker314314: model can be typed as int?

lib/Search/Provider.php
```diff
@@ -38,9 +43,11 @@ function search($query) {
 		$userId = \OCP\User::getUser();
 		$ownerView = new \OC\Files\View('/'. $userId . '/files');
 
+		$model = intval($this->config->getAppValue('facerecognition', 'model', AddDefaultFaceModel::DEFAULT_FACE_MODEL_ID));
+
 		$searchresults = array();
 
-		$results = $this->faceMapper->findAllNamedLike($userId, $query);
```
stalker314314: if this is last usage of findAllNamedLike, consider removing it completely (as there is no logic for accessing it anymore)

PR #51, opened 2018-11-15
Add exceptions to background tasks

stalker314314:
There was couple of todos for this. Basically, idea was that task throw
exceptions on errors and that main event loop for tasks handle that.
Some tasks are already throwing exceptions, some are throwing now,
but yes - control flow should be with exceptions, not with return values.

I also tried to fix scrutinizer errors from files that I already modified.

### DIFF
--- a/lib/BackgroundJob/BackgroundService.php
+++ b/lib/BackgroundJob/BackgroundService.php
@@ -95,25 +95,34 @@ class BackgroundService {
 			UnlockTask::class
 		];
 
-		// todo: implement bailing with exceptions
 		// Main logic to iterate over all tasks and executes them.
 		//
 		$startTime = time();
-		for ($i=0; $i < count($task_classes); $i++) {
+		for ($i=0, $task_classes_count = count($task_classes); $i < $task_classes_count; $i++) {
 			$task_class = $task_classes[$i];
 			$task = $this->application->getContainer()->query($task_class);
 			$this->context->logger->logInfo(sprintf("%d/%d - Executing task %s (%s)",
 				$i+1, count($task_classes), (new \ReflectionClass($task_class))->getShortName(), $task->description()));
 
-			$generator = $task->do($this->context);
-			foreach ($generator as $_) {
-				$currentTime = time();
-				if (($timeout > 0) && ($currentTime - $startTime > $timeout)) {
-					$this->context->logger->logInfo("Time out. Quitting...");
-					return;
-				}
+			try {
+				$generator = $task->do($this->context);
+				foreach ($generator as $_) {
+					$currentTime = time();
+					if (($timeout > 0) && ($currentTime - $startTime > $timeout)) {
+						$this->context->logger->logInfo("Time out. Quitting...");
+						return;
+					}
 
-				$this->context->logger->logDebug('yielding');
+					$this->context->logger->logDebug('yielding');
+				}
+			} catch (\Exception $e) {
+				// Any exception is fatal, and we should quit background job
+				//
+				$this->context->logger->logInfo("Error during background task execution");
+				$this->context->logger->logInfo("If error is not transient, this means that core component of face recognition is not working properly");
+				$this->context->logger->logInfo("and that quantity and quality of detected faces and person will be low or suboptimal.");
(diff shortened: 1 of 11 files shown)


lib/BackgroundJob/Tasks/CheckRequirementsTask.php
```diff
@@ -62,17 +62,18 @@ public function do(FaceRecognitionContext $context) {
 		$req = new Requirements($context->appManager, $model);
 
 		if (!$req->pdlibLoaded()) {
-			$this->logInfo('PDLib is not loaded. Cannot continue');
-			// todo: convert to exception
-			return;
+			$error_message = "PDLib is not loaded. Cannot continue";
+			$this->logInfo($error_message);
+			throw new \RuntimeException($error_message);
```
matiasdelellis: I do not think it should be an exception.. 
If you are checking something, should tell you if it is correct or not. The exceptions should be when you want to use something, and it fails unexpectedly.. Not when you are checking it ..

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
```diff
@@ -62,17 +62,18 @@ public function do(FaceRecognitionContext $context) {
...
-			$this->logInfo('Please contact administrator to change models you are using for face recognition');
-			$this->logInfo('or reinstall application. File an issue here if that doesn\'t help: https://github.com/matiasdelellis/facerecognition/issues');
-			// todo: convert to exception
-			return;
+			$error_message =
+				"Files of model with ID ' . $model . ' are not present in models/ directory.\n" .
+				"Please contact administrator to change models you are using for face recognition\n" .
+				"or reinstall application. File an issue here if that doesn\'t help: https://github.com/matiasdelellis/facerecognition/issues";
+			$this->logInfo($error_message);
+			throw new \RuntimeException($error_message);
```
matiasdelellis: Same here...
The exception must be if I try to use the model and fail, but not when corroborating if they exist ..

lib/BackgroundJob/BackgroundService.php
```diff
@@ -95,25 +95,34 @@ public function execute(int $timeout, bool $verbose, IUser $user = null) {
...
+						$this->context->logger->logInfo("Time out. Quitting...");
+						return;
+					}
 
-				$this->context->logger->logDebug('yielding');
+					$this->context->logger->logDebug('yielding');
+				}
+			} catch (\Exception $e) {
+				// Any exception is fatal, and we should quit background job
+				//
```
matiasdelellis: Handle exception here is correct, but should not fail when control a simple dependency, since it is a routine control...

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
stalker314314: So, what would be your proposal?
1. To have this `do` function implement bool return type that tells if we should continue or not? I need to double-check how to handle PHP generator functions and return types (as `do` is yielding and already returns generator object, I think I have to fiddle with something like [this](https://secure.php.net/manual/en/generator.getreturn.php))
2. Have our exception type, let's say `UnsatisfiedRequirementException extends LogicException`, and then we catch only `RuntimeException` in `BackgroundService::execute`

My understanding is that you want to avoid exception as control flow mechanism (routine control). And that is fine and I am OK with it, but I don't know what approach is best. Any other ideas except approach 1?

matiasdelellis: My doubt is rather conceptual. I'm thinking about it .. But I'll probably approve this proposal directly.

stalker314314: Check now, I am now returning `mixed` object from generators, both Generator and (at the end) bool. Had to be careful on return types in https://github.com/matiasdelellis/facerecognition/pull/51/files#diff-3eec1bf8e625d99e338e741f9c524b9cR112 but it works. If you sign off, let's commit (or you do it)

stalker314314:
I will optimistically merge this, as I have another PR that tests tasks and I need this change to test return from tasks. Please let me know if something needs to be changed and I can fix it later!

PR #53, opened 2018-11-17
Scrutinizer fix, licence in badge fix and adding composer, fixes #52

### LINKED ISSUE #52: Add composer.json
There are bunch of error in Scrutinizer that some import is missing and that they can be fixed if we have composer.json. I am talking about random stuff like [this](https://scrutinizer-ci.com/g/matiasdelellis/facerecognition/issues/master/files/lib/BackgroundJob/FaceRecognitionLogger.php?selectedLabels%5B0%5D=9&selectedSeverities%5B0%5D=10&orderField=path&order=asc&honorSelectedPaths=0). I noticed other apps actually have this file. Other than that, I do not see (currently) reason to have it, there is no dependencies for this project:) However, I guess it would be nice to have it...one day:)

stalker314314:
But it could also be called "how this thing worked at all":)


matiasdelellis:
> But it could also be called "how this thing worked at all":)

haha.. :man_facepalming:

composer.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "matiasdelellis/facerecognition",
+    "type": "project",
+    "license": "AGPLv3",
+    "authors": [
+        {
+            "name": "Matias Delellis"
```
matiasdelellis: Typo.. But don't worry.. 
'Matias De lellis'

stalker314314: ops, sorry, fixed:D

PR #54, opened 2018-11-21
Let's try this

stalker314314:
(will not push, just testing [this](https://scrutinizer-ci.com/docs/tools/php/php-analyzer/guides/migrate_to_new_php_analysis) )


stalker314314:
I will optimistically merge this, as I have another PR that tests tasks and I need this change to test return from tasks. Please let me know if something needs to be changed and I can fix it later!

PR #55, opened 2018-11-24
Adding test for tasks and introducing integration tests

stalker314314:
This commit adds some tests for tasks. Lock and Unlock could be tested
with unit tests, but AddMissingImagesTask couldnt, so I added integration
tests which are using proper database. Other tests are blocked until we can
install PDLib in system. Also added TESTING.md (to remind me how to run tests:)

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,30 @@
+# Running tests
+
+All tests should be easily executed with simple:
+```
+make test
+```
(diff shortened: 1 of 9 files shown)


matiasdelellis:
> Other tests are blocked until we can install PDLib in system

What do you need for this?. Is a debian or ubuntu package enough?.

I have a basic debian package, but I'm out of the office two weeks to fix it.

stalker314314:
Ubuntu is on Travis machines, so something that we can `sudo apt-get install` from some source? Anyhow, this is minor issue, it can wait next year also:)

PR #58, opened 2018-12-04
WIP, WIP, WIP..  Move main page no new db..

matiasdelellis:
Hi,
I only publish it so you can evaluate the clustering, but it does not have any extra functionality .. It only works to see the groups ..

 * About the user preferences pane, it is empty.
 * About the changes in the main application -That will be removed-, just add it to evaluate how to do it well..

Well, in my case..
imagen

In the main group, it confuses me with 9 people over 190 grouped faces. About a 4.74% error..

But in other cases ... the groups are ridiculous .. :sweat_smile: 
imagen

Tag my aunt in a photo .. and the rest does not match one .. haha


lib/Controller/PersonController.php
```diff
@@ -56,9 +62,18 @@ public function index() {
...
 			$cluster = [];
+			$faces = [];
+			$personFaces = $this->faceNewMapper->findFacesFromPerson($this->userId, $person->getId(), $model);
+			foreach ($personFaces as $personFace) {
+				$image = $this->imageMapper->find($this->userId, $personFace->getImage());
+				$face = [];
+				$face['id'] = $personFace->getId();
+				$face['file-id'] = $image->getFile();
+				$faces[] = $face;
+			}
```
matiasdelellis: It is not ideal so many embedded queries, but even adding more data, reduce the final query from 111Kb to 33Kb. :sweat_smile:

matiasdelellis:
Hi @stalker314314 
I just need to rename the groups here, and then I start to clean the branch, for a final merge.

p.s: There are millions of other things to add and improve, but this is the minimum and necessary. :wink:

stalker314314:
Sure, go ahead. I signed, so merge whatever branch and whenever you can!:)

PR #60, opened 2018-12-07
Drop old commands, routes, and unused controllers

matiasdelellis:
Still WIP, but open this PR to follow the results of the test..  :wink:


stalker314314:
Huh, big one (will be a lot of rebasing on my end:D). Not sure if we are ready for this "transition". I would love this to happen, don't get me wrong!, but not sure if this is "good enough" at this moment. If you say so, boss. I have no objection.

BTW, I still didn't turn on scheduled task to work (so there will be nothing "automatic" at this point)

matiasdelellis:
> will be a lot of rebasing on my end.

Sorry... :man_facepalming: :sweat_smile: 

> Not sure if we are ready for this "transition"

I have my doubts about the clustering method.. and I still prefer the use of fixed faces by the user and clustering around them. But the results are acceptable.. :wink: 

In any case can be migrated later just by adding a column 'user-aproved' to the face_recognition_faces table.

> but not sure if this is "good enough" at this moment.

We can discuss if it can be published, but about for the background task, it's much more robust than the ones I'm eliminating.. :grimacing: 

> BTW, I still didn't turn on scheduled task to work (so there will be nothing "automatic" at this point)

A cron task is more than enough.. :wink: 

Regards,
Matias.

PR #61, opened 2018-12-09
Dynamical image scaling, closes #59

### LINKED ISSUE #59: Decide on scaling image
Today, we scale image to 1024px (1024px is larger dimension!), with upsampling set to 0. This is mostly value that is set because it worked on my machine😄. Going forward, I wanted to see what is really this value. Higher value means more RAM is needed. Lower value means small faces will not be detected. So, I did some analysis and published it on my blog:
https://blog.kokanovic.org/performance-analysis-of-dlibs-cnn-face-detection/

@matiasdelellis - please read when you find time (it is lot of text😄). Basically, some questions for you:
* Are you OK to move content from blog post to wiki here?
* Are you OK to commit python code, output, produced images to repo here (let's say directory "analysis/" in root?)
* Are you OK with hardcoding upsampling to 0 and changing scaling based on RAM of user? I would go with something like this:
  * Read total physical memory on machine
  * If it is less than 2GB and we don't allow usage of Face Recognition app (I really cannot imagine how it can be usable with less than that:/ And even that is bare minimum)
  * Assume that free/available memory is 1/2 of total RAM memory (I would not go into inpecting `free -h` results (or amount of swap space) and use that in calculations, as it changes over time and we want consistent budget over time (unless user upgrades RAM physically😄)
  * Formula: `Scale [megapixels] = RAM[GB] / 1000` (so, 2000GB of RAM will scale to 2MP, 4000GB to 4MP...)

Again, I would not go into inspecting currently free memory, because scaling will be very dependent on current load on machine. Also, I would not go into some self-adaptive logic (for example, if we hit OOM, we try again with lower size...)

Your thoughts?

stalker314314:
This is implementation of out discussion in #59. It still missing some stuff, like:
* ability for user to manually choose quality (if I have 8GB, I don't want to use "only" 4GB, if I can tell 6GB)
* resiliency to OOM
* GPU calculation (if I have GPU, DLib do not care if I have 4GB or 8GB - this requires some changes in pdlib, I think)
* Windows/Mac/BSD do not have implementation to get system memory

I will create separate issues for all of these, but only "OOM ressiliency" is important for release, IMHO

### DIFF
--- a/lib/BackgroundJob/Tasks/CheckRequirementsTask.php
+++ b/lib/BackgroundJob/Tasks/CheckRequirementsTask.php
@@ -27,6 +27,7 @@ use OCP\IConfig;
 
 use OCA\FaceRecognition\BackgroundJob\FaceRecognitionBackgroundTask;
 use OCA\FaceRecognition\BackgroundJob\FaceRecognitionContext;
+use OCA\FaceRecognition\Helper\MemoryLimits;
 use OCA\FaceRecognition\Helper\Requirements;
 use OCA\FaceRecognition\Migration\AddDefaultFaceModel;
 
@@ -72,6 +73,28 @@ class CheckRequirementsTask extends FaceRecognitionBackgroundTask {
 				"Files of model with ID ' . $model . ' are not present in models/ directory.\n" .
 				"Please contact administrator to change models you are using for face recognition\n" .
 				"or reinstall application. File an issue here if that doesn\'t help: https://github.com/matiasdelellis/facerecognition/issues";
+			$this->logInfo($error_message);
+			return false;
+		}
+
+		$memory = MemoryLimits::getAvailableMemory();
+		if ($memory <= 0) {
+			// We cannot determine amount of memory to give to face recognition CNN.
+			// We will hardcode it here to 1GB, but plan is to expose this to user in future.
+			// TODO: allow user to choose "recognition quality", which will map to given memory.
+			// If user explicitely set something, we ignore getting memory from system.
+			$memory = 1024 * 1024 * 1024;
+		}
+
+		$context->propertyBag['memory'] = $memory;
+
+		if ($memory < 1024 * 1024 * 1024) {
+			$error_message =
+				"\n" .
+				"Seems that you have only " . intval($memory / (1024 * 1024)). "MB of memory given to PHP.\n" .
+				"Face recognition application requires at least 1 GB. You need to change your memory_limit in php.ini.\n" .
+				"Check https://secure.php.net/manual/en/ini.core.php#ini.memory-limit for details.\n" .
+				"If you already set this to unlimited, it seems your system is not having enough RAM memory.";
 			$this->logInfo($error_message);
 			return false;
 		}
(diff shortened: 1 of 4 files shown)


matiasdelellis:
p.s: About BSD, I would not worry. There is no way to ensure that you walk in all its derivatives.

stalker314314:
> We make pdlib report this compiled with optimizations, and cuda or not. But this does not ensure that it is being used or not.

Yes, of course, we need both checks if compiled with support and if CUDA can be initialized. If there is not support for that in DLib (and I think your reference just said there is no support for it:D), then...well...nothing we can do to be smarted on our side at this point:/ I created #64 and mentioned both checks.

Also opened #62, #63, #64 and #65 (I was highly productive in filing issues:D)

> Regarding the PR is approved, but -Although I am not the best example sweat_smile -, would be good to correct the reports of codacy.. wink

Regarding codacy, this is complete and utter sh$%^t. We should remove it, I was idiot for adding it. Scrutinizer was so awesome and actually pointed me to **real** issue, codacy is just...retarded. Maybe it is better for JS, but for PHP, it is completely retarded, in my 5-10 pull requests it pointed me to ZERO actionable stuff. C'mon, just look at how awesome these 7 issues that [Scrutinizer found](https://scrutinizer-ci.com/g/matiasdelellis/facerecognition/inspections/735d559b-fd51-4fce-b37f-8e51f39f0322/issues/) are, and look what codacy is finding above. They are not the same league of quality. I think you should remove Codacy? Plus codacy didn't found **any** of these 7 issues above in my first iteration! It is embarrassingly bad:)

matiasdelellis:
Hi,

> I created #64 and mentioned both checks.
> Also opened #62, #63, #64 and #65 (I was highly productive in filing issues:D)

:+1: 

> Regarding codacy, this is complete and utter sh$%^t.

I usually ignore and be quite flexible with style guides.. But for example, these reports I might have corrected.

> https://app.codacy.com/app/stalker314314/facerecognition/pullRequest?prid=2701855

```
- $memory = MemoryLimits::getAvailableMemory();
+ $memory = $this->getAvailableMemory();
```
You are in the same class, it is rare to use the static method.


And..

```
- if (php_uname("s") === "Linux") {
-   $linuxMemory = MemoryLimits::getTotalMemoryLinux();
-    ....
-    return intval($linuxMemory / 2);
- }
- else {
-     return -2;
-}
+ if (php_uname("s") !== "Linux") 
+     return -2;
+
+   $linuxMemory = MemoryLimits::getTotalMemoryLinux();
+    ....
+    return intval($linuxMemory / 2);
```

It is not the solution proposed of codacy, but it is better to return the error immediately, or in any case continue without following nesting code.. 
I usually program a lot in C. Therefore, if you think at the compiler level and optimizations, nesting code is a serious error. But we're in PHP, not C. So I do not worry. haha..

I repeat that I do not demand any changes. As you say, they are not errors, and it would be difficult for us to agree with the styles. :sweat_smile: 

I'm just saying that, although we prefer to ignore the reports, as I ignore all the ridiculous reports of double quote use in JS. it's good to have them.

Take them like that. Recommendations.. :wink:

PR #68, opened 2018-12-13
Add transtlations..

matiasdelellis:
Español to test that.. :smile: 

If you want to translate to some language:
1. make l10n_deps or make deps
2. mkdir translationfiles/es
3. cp translationfiles/templates/facerecognition.pot translationfiles/es/facerecognition.po
4. nano translationfiles/es/facerecognition.po
5. Translate.. . :wink: 
6. make update-translations

This generates the files l10n/es.js and l10n/es.json, and you could enjoy your translation.. :smile: 

Replace 'es' with language code from https://www.gnu.org/software/gettext/manual/html_node/Usual-Language-Codes.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,7 +1,7 @@
 build
-opt
 js/handlebars.js
 js/lozad.js
 vendor/
 models/1
 clover.xml
+translationtool.phar
--- a/l10n/es.js
+++ b/l10n/es.js
@@ -6,12 +6,12 @@ OC.L10N.register(
     "Please enter a name to rename the person" : "Por favor ingrese un nombre para renombrar la persona",
     "Rename" : "Renombrar",
     "Rename Person" : "Renombrar Persona",
-    "Cancel" : "Cancelae",
+    "Cancel" : "Cancelar",
     "Face Recognition" : "Reconocimiento Facial",
-    "A face recognition app" : "A aplicacion de reconocimiento facial",
+    "A face recognition app" : "Una aplicacion de reconocimiento facial",
     "Open Documentation" : "Abrir Documentación",
-    "Configuration information" : "Información de la configuracion",
+    "Configuration information" : "Información de la configuración",
     "Here you can see the photos of your friends that we have recognized" : "Aqui puedes ver las fotos de tus amigos que reconocimos",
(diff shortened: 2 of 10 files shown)


l10n/es.js
```diff
@@ -0,0 +1,17 @@
+OC.L10N.register(
+    "facerecognition",
+    {
+    "Persons" : "Personas",
+    "Looking for faces in this image..." : "Buscando por rostros en esta imagen...",
```
stalker314314: ... is usually ellipsis sign these days (…)

l10n/es.js
```diff
@@ -0,0 +1,17 @@
...
+    "Please enter a name to rename the person" : "Por favor ingrese un nombre para renombrar la persona",
+    "Rename" : "Renombrar",
+    "Rename Person" : "Renombrar Persona",
+    "Cancel" : "Cancelar",
+    "Face Recognition" : "Reconocimiento Facial",
+    "A face recognition app" : "Una aplicacion de reconocimiento facial",
+    "Open Documentation" : "Abrir Documentación",
+    "Configuration information" : "Información de la configuración",
+    "Here you can see the photos of your friends that we have recognized" : "Aqui puedes ver las fotos de tus amigos que reconocimos",
```
stalker314314: "the photos" -> "photos".

Also, I would refrain from "we" and use passive voice, something like "that are recognized/that have been recognized". #Consider

.gitignore
```diff
@@ -1,7 +1,7 @@
 build
-opt
 js/handlebars.js
 js/lozad.js
 vendor/
 models/1
 clover.xml
+translationtool.phar
```
stalker314314: I didn't use this before. Is this "standard" tooling in Nextcloud world? E.g. can we move to transifex with .po files. Asking because I know .po files are very powerful (and you can have multiple plurals, like Serbian language), but Transifex (which hosts most of NC and NC apps) have only 1 plural. Not sure if this is due to 1) Transifex being limited, or 2) other NC apps are using some less powerful format than .po

l10n/es.js
matiasdelellis: I still resist to unicode...  :sweat_smile:

PR #69, opened 2018-12-14
Fixing #67 and lowering memory limits to 3/4

### LINKED ISSUE #67: Drop database as test, and found at lest two bugs.
Drop all my faces tables as test and launch the background task. The analysis is done completely, but there is no clusters.

On log see these:
 
> A non well formed numeric value encountered at /var/www/html/nextcloud/apps/facerecognition/lib/Helper/MemoryLimits.php#105

https://github.com/matiasdelellis/facerecognition/blob/master/lib/Helper/MemoryLimits.php#L105

And:
> Division by zero at /var/www/html/nextcloud/apps/facerecognition/lib/BackgroundJob/Tasks/CreateClustersTask.php#167

https://github.com/matiasdelellis/facerecognition/blob/master/lib/BackgroundJob/Tasks/CreateClustersTask.php#L167

In the database now observed that:
 * oc_face_recognition_persons is empty.
 * oc_face_recognition_faces is empty.
 * oc_face_recognition_images has all the files, but with error 'std::bad_alloc'.

stalker314314:
This is trying to fix #67:
* fixed create cluster division by 0
* fixed wrong type for `Face::fromModel`
* lowered memory requirement to 3/4 of what was it before (1GB was enough for 1024x1024px, now it is enough "only" for 1024x768px)
* Upper bound for memory is 4GB (if user have more than that, it is ignore, it is not contributing to quality, just increases time to recognize faces)
* I don't think `returnBytes` was problem, because that warning (that you are seeing @matiasdelellis ) are just that - warnings, but code was more or less working (altough it is weird multplying int and string:). So, I *think* this is not what was causing problem for you, I think OOM & other stuff together was more likely. I did correct types in that function to fix warning (and because I am embrassed🙁)

Basically, I am not quite sure if this fixes everything, but please @matiasdelellis try it out when you find time with these fixes!

### DIFF
--- a/lib/BackgroundJob/Tasks/CheckRequirementsTask.php
+++ b/lib/BackgroundJob/Tasks/CheckRequirementsTask.php
@@ -84,6 +84,9 @@ class CheckRequirementsTask extends FaceRecognitionBackgroundTask {
 			// TODO: allow user to choose "recognition quality", which will map to given memory.
 			// If user explicitely set something, we ignore getting memory from system.
 			$memory = 1024 * 1024 * 1024;
+		} else {
+			// No point going above 4GB ("4GB should be enough for everyone")
+			$memory = min($memory, 4 * 1024 * 1024 * 1024);
 		}
 
 		$context->propertyBag['memory'] = $memory;
--- a/lib/BackgroundJob/Tasks/CreateClustersTask.php
+++ b/lib/BackgroundJob/Tasks/CreateClustersTask.php
@@ -164,7 +164,10 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 			// todo: 2 queries to get these 2 counts, can we do this smarter?
 			$imageCount = $this->imageMapper->countUserImages($userId, $modelId);
 			$imageProcessed = $this->imageMapper->countUserProcessedImages($userId, $modelId);
-			$percentImagesProcessed = $imageProcessed / floatval($imageCount);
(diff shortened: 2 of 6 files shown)


matiasdelellis:
Hi Branko,

> This is trying to fix #67:

Great .. It was what I expected ...

>     * fixed create cluster division by 0
> 
>     * fixed wrong type for `Face::fromModel`
:+1: 

>     * lowered memory requirement to 3/4 of what was it before (1GB was enough for 1024x1024px, now it is enough "only" for 1024x768px)
>     * Upper bound for memory is 4GB (if user have more than that, it is ignore, it is not contributing to quality, just increases time to recognize faces)

I just came thinking something similar. I'm glad you added it.. :wink: 

>     * I don't think `returnBytes` was problem, because that warning (that you are seeing @matiasdelellis ) are just that - warnings, but code was more or less working (altough it is weird multplying int and string:). So, I _think_ this is not what was causing problem for you, I think OOM & other stuff together was more likely. I did correct types in that function to fix warning (and because I am embrassedslightly_frowning_face)

Evidently it is not a problem and it still worked correctly, but obviously -Although it rejects what were previously characteristics of the language> PHP is getting more stricter, and we have to take it into account ... :disappointed_relieved: 
 
> Basically, I am not quite sure if this fixes everything, but please @matiasdelellis try it out when you find time with these fixes!

I approve because it looks good .. But I will repeat the test .. :wink:

PR #74, opened 2018-12-16
Fix task arguments

matiasdelellis:
Fix that the user argument did not work due the IUserManager injection was missing

On the other hand, before, if you want to pass the timeout, you must pass the user. Now you can pass them individually ..


stalker314314:
you are on fire these days, I am barely catching up with you:)

PR #76, opened 2018-12-17
Consider the is_procesed in view of the FileController

matiasdelellis:
Hi @stalker314314 
I'm already dizzy !, and I need your help .. :disappointed: 

The idea is, in the image view, consider is_processed to show better information about the status of the analysis on that photo..

But this commit results in:
> BadFunctionCallException: isProcessed is not a valid attribute

When do the query:
> https://github.com/matiasdelellis/facerecognition/compare/restore-progress?expand=1#diff-5dc310fdcd7b76f8112d2d7b4ed94866R64


stalker314314:
I could swear this is related to how mapper is getting attributes (throught reflection). Consider renaming attribute to "isProcessed" (instead of "is_processed"). I had to do this for creationTime in face. If this doesn't work, I will take this branch and take a look!

matiasdelellis:
D'Oh!. We are wrong with the conventions ...  :sweat_smile: 

> https://docs.nextcloud.com/server/14/developer_manual/app/database.html#entities

With the last commit it works.. :wink: 

There are many others that is wrong, but I leave it for another pull request.. 

Now I continue with the improvements of the files tab view-.

matiasdelellis:
Now it has four 'views'.

1. Looking for faces in this image…
2. This image is not yet analyzed
3. No people found
4. *List of peoples*

Again, I hope opinion of the messages .. :wink:

lib/Controller/FileController.php
```diff
@@ -56,6 +63,13 @@ public function getPersonsFromPath(string $fullpath) {
 		$fileId = $userFolder->get($fullpath)->getId();
 
 		$resp = array();
+
+		$image = $this->imageMapper->findFromFile($this->userId, $fileId);
+		$resp['image_id'] = $image ? $image->getId() : 0;
+		$resp['is_processed'] = $image ? $image->getIsProcessed() : 0;
+		$resp['error'] = $image ? $image->getError() : null;
+
+		$aPersons = array();
```
stalker314314: you don't need this variable. You can just use `$resp['persons']`?

stalker314314:
watch out for scrutinizer, always have something smart to say:)
https://scrutinizer-ci.com/g/matiasdelellis/facerecognition/inspections/b7d252b5-6a23-4f15-a328-885f79e3fbb1/issues/files/lib/Db/Image.php?status=new&orderField=path&order=asc&honorSelectedPaths=0
(actually, smart enough to suggest name of variable?:)

lib/Controller/FileController.php
matiasdelellis: Something like that?
I have to force the creation of the *$resp['persons']*  to verify that it is empty..

matiasdelellis:
> (actually, smart enough to suggest name of variable?:)

haha.. Done..

lib/Controller/FileController.php
stalker314314: exactly what I had in mind! thanks!

PR #79, opened 2018-12-18
Update min version to NC 14 and PHP 7.0

matiasdelellis:
😉

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -19,8 +19,8 @@
 		<filesystem/>
 	</types>
 	<dependencies>
-		<php min-version="5.6"/>
-		<nextcloud min-version="11" max-version="14" />
+		<php min-version="7.0"/>
+		<nextcloud min-version="14" max-version="15" />
 	</dependencies>
 	<repair-steps>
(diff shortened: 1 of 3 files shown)


appinfo/info.xml
```diff
@@ -19,8 +19,8 @@
 		<filesystem/>
 	</types>
 	<dependencies>
-		<php min-version="5.6"/>
-		<nextcloud min-version="11" max-version="14" />
+		<php min-version="7.0"/>
+		<nextcloud min-version="14" max-version="15" />
```
stalker314314: there are 2 other things to be done, if we are going to 14:
* remove QBMapper (it is not needed anymore)
* composer.json should be updated (I hope this will improve travis and those stupid errors with `select` of multiple parameters)

I mean, none of these 2 are a must for this checkin, but we should not forget them

matiasdelellis: > composer.json should be updated (I hope this will improve travis and those stupid errors with select of multiple parameters)

Done.. waiting for the result ..

> remove QBMapper (it is not needed anymore)

From where?

matiasdelellis: QBMapper was introduced on 14.. in any case we must migrate to QBMapper.. Ok?

PR #81, opened 2018-12-19
Removing check for FULL_IMAGE_SCAN_DONE_KEY, fixes #77

### LINKED ISSUE #77: There is no clustering until I analyze all the photos of all the users.
I just want to enable a user for tests but until I analyze all the users, grouping is not done... :disappointed:

On the other hand the task CreateClustersTask is executed before ImageProcessingTask, therefore, all the photos must be analyzed, and then launch it again to do the grouping.. :wink:

stalker314314:


matiasdelellis:
About bug #77 , I think it's solved.. 
So, aprove that.. :wink: 

Then, based on all this discussion, I think that my next PR will be to add a configuration per user, to enable or disable scanning/clustering.

PR #84, opened 2018-12-21
WIP: Add check box to enable analysis by user..

matiasdelellis:
WIP .. I just started ...

The idea that like other similar applications, it is disabled by default, and that each user enables it.

A lot of people do not like to be told:
> we are analyzing your photos, to find your friends..

:sweat_smile: 

Because of them, the popularity of nextcloud ?? haha

### DIFF
--- a/lib/Controller/SettingController.php
+++ b/lib/Controller/SettingController.php
@@ -33,7 +33,7 @@ class SettingController extends Controller {
 	                             IConfig      $config,
 	                             FaceMapper   $faceMapper,
 	                             ImageMapper  $imageMapper,
-	                             PersonMapper $personMapper
+	                             PersonMapper $personMapper,
 	                             $userId)
 	{
 		parent::__construct($appName, $request);
--- a/lib/Controller/SettingController.php
+++ b/lib/Controller/SettingController.php
@@ -93,17 +93,21 @@ class SettingController extends Controller {
 	protected function setEnableUser (string $userId, $enabled): bool {
 		$success = false;
 		$this->config->setUserValue($userId, $this->appName, 'enabled', $enabled);
-		if ($enabled == 'false') {
+		if ($enabled === 'false') {
 			// TODO: Invalidate images and remove user info here???
-			//$this->faceMapper->resetUser($this->userId);
-			//$this->personMapper->resetUser($this->userId);
-			//$this->imageMapper->resetUser($this->userId);
+			//$this->resetUser ($userId);
 			$success = true;
 		}
 		else {
 			$success = true;
 		}
 		return $success;
+	}
+
+	protected function resetUser (string $userId) {
+		$this->faceMapper->resetUser($userId);
+		$this->personMapper->resetUser($userId);
+		$this->imageMapper->resetUser($userId);
 	}
 
 }
--- a/lib/Controller/SettingController.php
+++ b/lib/Controller/SettingController.php
@@ -31,8 +31,18 @@ class SettingController extends Controller {
 	 * @throws \OCP\PreConditionNotMetException
 	 */
 	public function setValue($type, $value) {
-		$this->config->setUserValue($this->userId, $this->appName, $type, $value);
-		return new JSONResponse(array('success' => 'true'));
+		$success = false;
+		switch ($type) {
+			case 'enabled':
+				$success = $this->setEnableUser($this->userId, $value);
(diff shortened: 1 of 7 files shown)


matiasdelellis:
> I guess there will be backend pieces in this PR?

From my side I will add, that the background task, select the users according to this user settings.

> I guess former, but we need additional button to wipe everything

When you click, I was thinking about adding a confirmation request, and if you agree to delete the information directly.

However, now that you mention it, this can bring problems if the task is running..

Well .. I accept ideas .. :wink:

lib/Controller/SettingController.php
```diff
@@ -58,4 +68,17 @@ public function getValue($type) {
 		return $response;
 	}
 
+	protected function setEnableUser (string $userId, $enabled): bool {
+		$success = false;
+		$this->config->setUserValue($this->userId, $this->appName, 'enabled', $enabled);
+		if ($enabled == 'false') {
+			// TODO: Invalidate images and remove user info here???
```
matiasdelellis: @stalker314314 
what you think? Is it a good place to delete user data here?

stalker314314: TBH, I would say no. I would split this to 2 separate API calls, one to simply change value, one to do deletion (deletion will be used on some other places, I think (when resetting, when changing model, when admin delete all users data...)).

On the client side, when user uncheck this, I would pop up additional question, like "do you want to keep already detected faces?" and do additional "delete" call if user says no. This way, we get both "pause" and "wipe" semantics.

If you add resetting here in setEnableUser, you are merging them unnecessarly. My thinking, do what you think makes more sense:)

matiasdelellis: > TBH, I would say no. I would split this to 2 separate API calls, one to simply change value, one to do deletion

I think it adds complexity and does not contribute as much.. In this instance, it will remain this way ..
But i want to warn, and ask for confirmation from the user before deleting the data..  But not a second check. If you disable it, you must delete the data ..

> (deletion will be used on some other places, I think (when resetting, when changing model, when admin delete all users data...)).

For this, I will separate the method that performs the cleaning, but it will stay inside the same controller.
If you change the model, it can be done in the same controller, since this controller also can handle admin settings..

> If you add resetting here in setEnableUser, you are merging them unnecessarly. My thinking, do what you think makes more sense:)

To the people that I think with this commit, they do not want to leave the data in the server .. :wink:

stalker314314: Ok, it was just a proposal:) I understand your points, makes sense, go with it!;)

lib/Controller/SettingController.php
```diff
@@ -0,0 +1,113 @@
...
+	protected function setEnableUser (string $userId, $enabled): bool {
+		$success = false;
+		$this->config->setUserValue($userId, $this->appName, 'enabled', $enabled);
+		if ($enabled === 'false') {
+			// TODO: Invalidate images and remove user info here???
+			//$this->resetUser ($userId);
+			$success = true;
+		}
+		else {
+			$success = true;
```
stalker314314: I just remembered it now:) When going off->on, you should reset `FULL_IMAGE_SCAN_DONE_KEY`, so all images that were uploaded in the meantime are rescaned. Since this is highly inefficient (instance-level reset instead of user-level reset), I created this issue on me - #85.

You don't have to implement this resetting of `FULL_IMAGE_SCAN_DONE_KEY` now (in this PR), I cannot blame you:), it will be really ineffiecent:) I can add reset in this code later, once I implement #85.

lib/Db/FaceMapper.php
```diff
@@ -141,4 +141,16 @@ public function removeFaces(int $imageId) {
...
 
+	/**
+	 * Drop all Faces from that user.
+	 *
+	 * @param string $userId user to drop from table.
+	 */
+	public function resetUser(string $userId) {
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->getTableName())
+			->where($qb->expr()->eq('user', $qb->createNamedParameter($userId)))
```
matiasdelellis: The table face has no user column !!!!!!! :man_facepalming:

matiasdelellis:
Just rebase on master to continue with this tomorow.. :grimacing:

matiasdelellis:
Rebase again... :sweat_smile:

PR #87, opened 2018-12-26
Ignore small images (parametarized), fixes #65

### LINKED ISSUE #65: Do not do face recognition for small images
I got bunch of small files in my instance (from random programs, games, books, backups, I don't even know from where:)). They are not in my "main" directory of photos. I do not want recognition on them, because it will take needless time, needless DB space and will add noise to results. I think proper solution might be having whitelist/blacklist of folders per user, but it adds another layers of complexity. I think what we can do now is to ignore small files. Let's say - all files:
* with less than 0.5MP (megapixels), **or**
* width < 500px, **or**
* height < 500px
are most likely not coming from digital camera and we should be ignoring them. This should be configurable though, so advanced users can set this config.

stalker314314:


matiasdelellis:
Great..
If it is configurable, I have no objection..

PR #88, opened 2018-12-26
Split full image scans to be per user, fixes #85

### LINKED ISSUE #85: Rewrite FULL_IMAGE_SCAN_DONE_KEY to be per-user, not per-instance
When we reset `FULL_IMAGE_SCAN_DONE_KEY`, there is no reason to do full scan of **all** images of **all** users. Only affected users' images should be rescaned. This will also help logic from PR https://github.com/matiasdelellis/facerecognition/pull/84 where going off->on should trigger rescan.

stalker314314:
Here is what we discussed, this should help us get far less unnecessary work


matiasdelellis:
Ok.. 
It looks good, but I will not have time to try it. I trust in you.. :wink:

PR #90, opened 2018-12-29
Implement stale image removal task, fixes #37

### LINKED ISSUE #37: Add task to remove images that do not exist physically, but are in DB

stalker314314:
Adds new task to remove images that either:
* exist in database, but do not exist on disk (this can happen because of bugs in Watcher, in testing, in manual fiddling to filesystem...)
* are under .nomedia because user added that file to it

This also fixes some small issues found along the way.

I learned (hard way:smiley:) that SQLite cannot have table aliases in UPDATE clauses (https://travis-ci.org/matiasdelellis/facerecognition/jobs/473327871). Change in PersonMapper.php is because of that:)

I think this is all from me for this year;) Let's release something next year!:)

### DIFF
--- a/lib/Watcher.php
+++ b/lib/Watcher.php
@@ -33,6 +33,7 @@ use OCP\ILogger;
 use OCP\IUserManager;
 
 use OCA\FaceRecognition\BackgroundJob\Tasks\AddMissingImagesTask;
+use OCA\FaceRecognition\BackgroundJob\Tasks\StaleImagesRemovalTask;
 use OCA\FaceRecognition\Db\Face;
 use OCA\FaceRecognition\Db\Image;
 use OCA\FaceRecognition\Db\FaceMapper;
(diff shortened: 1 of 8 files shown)


lib/Watcher.php
```diff
@@ -110,13 +111,18 @@ public function postWrite(Node $node) {
 			return;
 		}
 
-		// todo: issue #37 - if we detect .nomedia written, reset some global flag such that RemoveMissingImagesTask is triggered.
+		$owner = $node->getOwner()->getUid();
 
-		if (!Requirements::isImageTypeSupported($node->getMimeType())) {
+		if ($node->getName() === '.nomedia') {
+			// If user added this file, it means all images in this and all child directories should be removed.
+			// Instead of doing that here, it's better to just add flag that image removal should be done.
```
matiasdelellis: Interesting solution... =)

Issue #97, opened 2019-01-06
No faces has been recognized - memory settings


SlavikCA:
I ran face recognition on few of my photos here:
https://cloud.aidcim.org/index.php/apps/gallery/s/6NFCGmAkoZL6Pc5

That's on
- Ubuntu 16.04 with 6GB RAM
- NextCloud 15.0

The job completes:
```
sudo -u www-data php occ -v face:background_job -u ****
1/8 - Executing task CheckRequirementsTask (Check all requirements)
2/8 - Executing task CheckCronTask (Check that service is started from either cron or from command)
3/8 - Executing task LockTask (Acquire lock so that only one background task can run)
4/8 - Executing task CreateClustersTask (Create new persons or update existing persons)
        Skipping cluster creation, not enough data (yet) collected. For cluster creation, you need either one of the following:
        * have 1000 faces already processed (you have 0),
        * have 100 images (you have 67),
        * or you need to have 95% of you images processed (you have 0.00%)
5/8 - Executing task AddMissingImagesTask (Crawl for missing images for each user and insert them in DB)
        Finding missing images for user google-106379083885103545634
        Found /google-106379083885103545634/files/Photos/05 Penasco/IMG_20180530_090646.jpg
        Found /google-106379083885103545634/files/Photos/05 Penasco/IMG_20180531_104813.jpg
        Found /google-106379083885103545634/files/Photos/05 Penasco/IMG_20180531_104832.jpg

...

yielding
        Processing image /var/www/nextcloud_data/google-106379083885103545634/files/Photos/05 Penasco/IMG_20180602_175238.jpg
yielding
        Processing image /var/www/nextcloud_data/google-106379083885103545634/files/Photos/05 Penasco/IMG_20180602_104556.jpg
yielding
        Processing image /var/www/nextcloud_data/google-106379083885103545634/files/Photos/05 Penasco/IMG_20180602_160705.jpg
8/8 - Executing task UnlockTask (Release obtained lock)
```

But the "Settings => Face Recognition" page still empty:
```
Your friends have not been recognized yet
Please, be patient
```

When I run it again, the job gets completed in a second:
```
sudo -u www-data php occ -v face:background_job -u google-106379083885103545634
1/8 - Executing task CheckRequirementsTask (Check all requirements)
2/8 - Executing task CheckCronTask (Check that service is started from either cron or from command)
3/8 - Executing task LockTask (Acquire lock so that only one background task can run)
[...]

PR #99, opened 2019-01-09
Adding more tracing when detecting memory limits

stalker314314:
Adding more tracing to see if we can understand issue #97 better

@matiasdelellis  Hope you don't mind, but I will push this directly


stalker314314:
Again @matiasdelellis hope you don't mind

Issue #101, opened 2019-01-09
OCP\AppFramework\Db\MultipleObjectsReturnedException


SlavikCA:
After somewhat successfully (according to log) scanning ~100+ photos, 
consequent scanning fails.

```
root@ubhome:/var/www/nextcloud/apps/facerecognition# sudo -u www-data php /var/www/nextcloud/occ -vvv face:background_job -vvv -u google-*****
1/8 - Executing task CheckRequirementsTask (Check all requirements)
        Found 4096 MB available to PHP.
2/8 - Executing task CheckCronTask (Check that service is started from either cron or from command)
3/8 - Executing task LockTask (Acquire lock so that only one background task can run)
4/8 - Executing task CreateClustersTask (Create new persons or update existing persons)
        Found 3 faces without associated persons for user google-106379083885103545634 and model 1
Error during background task execution
If error is not transient, this means that core component of face recognition is not working properly
and that quantity and quality of detected faces and person will be low or suboptimal.
You probably want to file an issue (please include exception below) to: https://github.com/matiasdelellis/facerecognition/issues

In QBMapper.php line 223:

  [OCP\AppFramework\Db\MultipleObjectsReturnedException]
  Did not expect more than one result when executing: query "SELECT `f`.`id`, `f`.`creation_time` FROM `*PREFIX*face_recognition_faces` `f` INNER JOIN `*PREFIX*face_recognition_images` `i`
   ON `f`.`image` = `i`.`id` WHERE (`user` = :dcValue1) AND (`model` = :dcValue2) AND (`person` IS NULL)";


Exception trace:
 OCP\AppFramework\Db\QBMapper->findOneQuery() at /var/www/nextcloud/lib/public/AppFramework/Db/QBMapper.php:287
 OCP\AppFramework\Db\QBMapper->findEntity() at /var/www/nextcloud/apps/facerecognition/lib/Db/FaceMapper.php:79
 OCA\FaceRecognition\Db\FaceMapper->getOldestCreatedFaceWithoutPerson() at /var/www/nextcloud/apps/facerecognition/lib/BackgroundJob/Tasks/CreateClustersTask.php:133
 OCA\FaceRecognition\BackgroundJob\Tasks\CreateClustersTask->createClusterIfNeeded() at /var/www/nextcloud/apps/facerecognition/lib/BackgroundJob/Tasks/CreateClustersTask.php:98
[...]

PR #103, opened 2019-01-12
Fixing #101 where multiple faces are returned

stalker314314:


stalker314314:
Again, @matiasdelellis - add this to pile of stuff to skim over when you are back:)

PR #108, opened 2019-01-26
Fix for non-existing image

stalker314314:
If image doesn't exist on disk, but do exist in Nextcloud filecache,
we don't see it, and we don't want to add it to our database
(by-design).
So, we signal that processing context is null, but code had bug
(regression actually) where it did not handle null case. This fixes it.

### DIFF
--- a/lib/BackgroundJob/Tasks/ImageProcessingTask.php
+++ b/lib/BackgroundJob/Tasks/ImageProcessingTask.php
@@ -162,9 +162,11 @@ class ImageProcessingTask extends FaceRecognitionBackgroundTask {
 					$this->populateDescriptors($fld, $fr, $imageProcessingContext);
 				}
 
-				$endMillis = round(microtime(true) * 1000);
-				$duration = max($endMillis - $startMillis, 0);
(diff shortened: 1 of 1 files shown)


lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -162,9 +162,11 @@ public function execute(FaceRecognitionContext $context) {
 					$this->populateDescriptors($fld, $fr, $imageProcessingContext);
 				}
 
-				$endMillis = round(microtime(true) * 1000);
-				$duration = max($endMillis - $startMillis, 0);
-				$this->imageMapper->imageProcessed($image, $imageProcessingContext->getFaces(), $duration);
```
matiasdelellis: I generally prefer something like:

> if ($imageProcessingContext === null) 
>    return true;

But is ok.. :wink:

stalker314314: Sorry about this, I also like that approach, neutralizes excessive indentation and makes code more readable. Sometimes, though, people need to point me in, khm, khm... right direction:smile: Fixed in second iteration!

PR #110, opened 2019-01-26
Adding support to remove data from deleted users, fixes #32

### LINKED ISSUE #32: Cleanup after user is deleted
When user is deleted, we should remove all persons, images and faces associated with this user. This can be implemented either as a hook or part of background job. I guess hook will block more, but we can detect issues better with this approach:)

stalker314314:
If user is deleted from Nextcloud, all files from that users are also deleted. However, we don't have callback for that, and we end up with bunch of images/faces/persons in database which do not exist anymore, either in filecache or physically, on disk. Since those stuff can pile up, best is to clean them from time to time, and we use another task for that.

Tests also added.

### DIFF
--- a/lib/Watcher.php
+++ b/lib/Watcher.php
@@ -82,15 +82,14 @@ class Watcher {
 	 * @param PersonMapper $personMapper
 	 * @param FaceManagementService $faceManagementService
 	 */
-	public function __construct(
-								IConfig               $config,
-								ILogger               $logger,
-								IDBConnection         $connection,
-								IUserManager          $userManager,
-								FaceMapper            $faceMapper,
-								ImageMapper           $imageMapper,
-								PersonMapper          $personMapper,
-								FaceManagementService $faceManagementService)
+	public function __construct(IConfig               $config,
+	                            ILogger               $logger,
+	                            IDBConnection         $connection,
+	                            IUserManager          $userManager,
+	                            FaceMapper            $faceMapper,
+	                            ImageMapper           $imageMapper,
+	                            PersonMapper          $personMapper,
+	                            FaceManagementService $faceManagementService)
 	{
 		$this->config = $config;
 		$this->logger = $logger;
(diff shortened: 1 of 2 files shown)


stalker314314:
> You have to use hooks for it..

Not sure I understand, can you clarify? I already explained why hooks are not an option in associated issue #32. Do you mean that or sođe other hooks?

stalker314314:
OK, completely changed this pull request! Removed everything and replaced with hook. No tests, though (test for reset already exist). BTW, you can easily try it out with simple `./occ user:add` and `./occ user:delete` commands.

P.S. I was lucky to extract reset functionality into separate commit (https://github.com/matiasdelellis/facerecognition/commit/c4d1b2fdd24af8a451f5277bd3dc0bbcf1fb4c1a), so stuff here were easy to refactor.

lib/Watcher.php
```diff
@@ -75,14 +80,17 @@ class Watcher {
...
+	 * @param FaceManagementService $faceManagementService
 	 */
-	public function __construct(IConfig       $config,
-	                            ILogger       $logger,
-	                            IDBConnection $connection,
-	                            IUserManager  $userManager,
-	                            FaceMapper    $faceMapper,
-	                            ImageMapper   $imageMapper,
-	                            PersonMapper  $personMapper)
+	public function __construct(
```
matiasdelellis: I told you I would never demand changes styles.. Just comment.. :wink: 
The way to avoid these bad tabulations, is to use a mix of tab and spaces .. :sweat_smile:

matiasdelellis:
Just to be sure, when you were talking about business logic, you referring to using a single command that executes all the tasks?

lib/Watcher.php
stalker314314: fixed in version 2

stalker314314:
> Just to be sure, when you were talking about business logic, you referring to using a single command that executes all the tasks?

Yeah, in the sense "all of background processing". I think of it as all classes under Tasks. But, ignore my comments about "grouping business logic together", business logic is in Watcher.php too...

PR #111, opened 2019-01-26
Improve needless download of translationtool

stalker314314:
I already made mistake once with multiple translationtool.phar.N files:)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,7 +51,9 @@ js/lozad.js:
 
 javascript_deps: js/handlebars.js js/lozad.js
 
-l10n_deps:
+l10n_deps: translationtool.phar
+
+translationtool.phar:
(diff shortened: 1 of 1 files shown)


matiasdelellis:
Mm.. 
This change does not improve the multiple download.. :thinking:

matiasdelellis:
I not test it.. :disappointed: 
But you are just adding a new rule, and you put it as a dependency to the rule where you got it before..

Therefore, it is executed in the same way as before.. :sweat_smile: 

The solution is to use 'wget -o' instead of just using 'wget' to overwrite the file.. :wink:

matiasdelellis:
> I am adding target which has name == name of the file, and make will not execute that target if it exists already.

Ohh.. You're right ... :sweat_smile:

> (this is where all my knowledge of Makefiles comes from, I literally don't know anything else)

When you have to see autotools, you will escape scared.. haha..

Issue #92, opened 2019-01-01
Face recognition app is not showing any faces


stalker314314:
I have Nextcloud 14. Latest version from master of face recognition.

So, I am visiting `https://[site]/index.php/settings/user/facerecognition`. After some loading, this is what I get:
![slika](https://user-images.githubusercontent.com/188751/50572522-6751c100-0dba-11e9-997c-37dd3501ecdd.png)

When I hover over thumbnails, I get this link `https://[site]/f/35494` and when I click on that thumbnail, I get this:
![slika](https://user-images.githubusercontent.com/188751/50572578-4ccc1780-0dbb-11e9-93b4-afd4066c11ee.png)

In network tab (Firefox web developer tools), I see bunch of requests going to `https://[site]/apps/facerecognition/face/198/thumb/50`, but they all finish with 404.

Basically, what I need to do/tweak/install to see thumbnails?:)

matiasdelellis:
D'Oh!. Take it..

matiasdelellis:
Hi @stalker314314 @SlavikCA 

> Basically, what I need to do/tweak/install to see thumbnails?:)

'htaccess.RewriteBase' => '/',

But it's still a bug ... it should work with the full path ..

These days I start working again. :wink: 

Thank for all..

PR #112, opened 2019-01-29
Fix thumbnails when is not configured the clean url. Issue #92

matiasdelellis:
Hi @stalker314314 @SlavikCA 

With this small change everything should work correctly. However when I can, I should improve URL generation.. 

Please test.. :wink:


stalker314314:
I can confirm it works now on my "prod" machine! (while previously it didn't!)

PR #116, opened 2019-02-07
Remove postmigration step as not needed, closes #12

### LINKED ISSUE #12: Can we just delete AddDefaultFaceModel post-migration steps
We currently have `AddDefaultFaceModel` post-migration step which only does one thing - adds default model with ID=1. Maybe it can be part of installation step. What if we delete app and install it again?

stalker314314:


stalker314314:
sending PR as FYI, will merge without approval (hope you don't mind, change is trivial:D)

Issue #95, opened 2019-01-06
Personal => Face Recognition page result in error: the Uncaught EvalError: Refused to evaluate a string as JavaScript


SlavikCA:
I just installed the app manually on my system:
- Ubuntu 16.04
- Nextcloud 15

And when opening `Personal => Face Recognition` in the browser (Chrome 71), I see this error in the console:
```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' 
is not an allowed source of script in the following Content Security Policy directive:
"script-src 'nonce-d0FNdFhKT3pqc2F4NHRsWWV0UERhclVPb3FHYVdyNmVOdWE4U1g3N2NBaz06aGxWQUphRGR2NEhEcHZJTU1iaUdDUFpoODVueWFkZmtYYmFNY0N5Q0NWcz0='".

    at JavaScriptCompiler.Function (<anonymous>)
    at JavaScriptCompiler.createFunctionContext (handlebars.js?v=27e5d68a-4:3562)
    at JavaScriptCompiler.compile (handlebars.js?v=27e5d68a-4:3450)
    at JavaScriptCompiler.compileChildren (handlebars.js?v=27e5d68a-4:4102)
    at JavaScriptCompiler.compile (handlebars.js?v=27e5d68a-4:3405)
    at JavaScriptCompiler.compileChildren (handlebars.js?v=27e5d68a-4:4102)
    at JavaScriptCompiler.compile (handlebars.js?v=27e5d68a-4:3405)
    at JavaScriptCompiler.compileChildren (handlebars.js?v=27e5d68a-4:4102)
    at JavaScriptCompiler.compile (handlebars.js?v=27e5d68a-4:3405)
    at compileInput (handlebars.js?v=27e5d68a-4:3249)
```

And way to fix it?

SlavikCA:
ok, this error has nothing to do with Face Recognition.

It was caused by some HTTP headers.

Fixed by adding this to Apache config in .htaccess:
```
    Header set Content-Security-Policy "default-src 'self';"
    Header set Content-Security-Policy "style-src 'self' 'unsafe-inline';"
```

That's taken from the forum here:  
https://help.nextcloud.com/t/nextcloud-13-0-4-doesnt-work-with-chrome-and-firefox-works-only-with-ie/33466/8

svolgmann:
I still stuck in the same problem with
- Nextcloud 15.0.2
- facerecognition 6e5dc95 

[...]

PR #118, opened 2019-02-12
Move to precompiled handlebars templates.

matiasdelellis:
In Nextcloud 15 the default Content Security Policy disallows unsafe
eval expressions, so Handlebars templates can no longer be compiled at
runtime.

Fix issue #95

### DIFF
--- a/lib/Controller/FaceController.php
+++ b/lib/Controller/FaceController.php
@@ -57,6 +57,7 @@ class FaceController extends Controller {
 		$img = new OCP_Image();
 		$fileName = $ownerView->getLocalFile($path);
 		$img->loadFromFile($fileName);
+		$img->fixOrientation();
 
 		$x = $face->getLeft ();
 		$y = $face->getTop ();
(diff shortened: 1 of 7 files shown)


matiasdelellis:
```
npm install handlebars -g
make js-templates
```

Hi @SlavikCA,
Test in NC15, I think this is enough, :smile: 

Hi @stalker314314 
Check that there is no repression in NC 14. :sweat_smile:

lib/Controller/FaceController.php
```diff
@@ -57,6 +57,7 @@ public function getThumb ($id, $size) {
 		$img = new OCP_Image();
 		$fileName = $ownerView->getLocalFile($path);
 		$img->loadFromFile($fileName);
+		$img->fixOrientation();
```
stalker314314: sweet:)

stalker314314:
> Check that there is no repression in NC 14.

Sorry, in a week or two, currently OOF. Proceed as if I checked:)

lib/Controller/FaceController.php
matiasdelellis: D'Oh!. 
I mixed that change that i was testing.. 😅
If confirm that fix nc15, I merge it like this.😌

Issue #119, opened 2019-02-17
TypeError: Handlebars.templates is undefined


stalker314314:
This might be continuation of #95.
After I merged https://github.com/matiasdelellis/facerecognition/commit/cb50dfd49d95e0ef83487e579c15614ce4327fc3, facerecognition page is blank. Here is output from firefox (I can translate if needed):
![slika](https://user-images.githubusercontent.com/188751/52911439-aae58900-329b-11e9-8a31-feaa68e05d9c.png)
and from edge:
![slika](https://user-images.githubusercontent.com/188751/52911418-75d93680-329b-11e9-819f-c340186648f0.png)

I did `make` before copying to my server, but [`js-templates` rule](https://github.com/matiasdelellis/facerecognition/commit/cb50dfd49d95e0ef83487e579c15614ce4327fc3#diff-b67911656ef5d18c4ae36cb6741b7965R54) was not part of default `make`? Also, if I run `make js-templates` rule manually, it is broken as I don't have `handlebars` executable.

This also doesn't work on my dev machine. Seems more likely that I am missing some dependency than coding issue? Looking at [this .gitignore](https://github.com/matiasdelellis/facerecognition/commit/cb50dfd49d95e0ef83487e579c15614ce4327fc3#diff-a084b794bc0759e7a6b77810e01874f2R3), I am missing `js/templates.js`?:
```
$:~/server/apps/facerecognition$ ls js -R
js:
admin.js  files-plugin.js  files-tabview.js  handlebars.js  lozad.js  personal.js  templates

js/templates:
personal.handlebars
```

matiasdelellis:
Hi branko,
https://github.com/matiasdelellis/facerecognition/pull/118#issuecomment-462787351

I did not do it as dependence, because it's more like a build.. but the executable, I do not want to install it from there, but at least we should warn of that dependence, and that we must execute it

matiasdelellis:
Hi again @stalker314314 ,

Check last PR.. :wink: 

p.s: By the way, a pleasure to know about you again!

stalker314314:
I agree "handlebars" must be present and we should only warn (same as with `wget`, `bzip2`...) and not install it. But it should be automatic, I think - I run `make` and I get `js/templates.js`, right?

stalker314314:
this is fixed:)

PR #120, opened 2019-02-17
Add an rule to check handlebars executable which is a dependency to b…

matiasdelellis:
…uild the application.

It is used to verify that it exists and remember that you must compile the templates.

Issue #119


Makefile
```diff
@@ -26,6 +28,18 @@ else
 	composer update --prefer-dist
 endif
 
+handlebars:
+ifeq (,$(handlebars))
+	@echo "No handlebars command available, remember to install this"
+	@echo " # npm install handlebars -g"
+	@echo "Then you must compile the templates"
```
stalker314314: I would not go into explaining to compile templates. Imagine users do first time `make` and this blows up and he thinks he should execute `make js-templates`:) Consider removing it or just telling "Then run make again" or "Then restart your command"...

Makefile
```diff
@@ -59,7 +73,7 @@ l10n_deps: translationtool.phar
 translationtool.phar:
 	wget https://github.com/nextcloud/docker-ci/raw/master/translations/translationtool/translationtool.phar -O translationtool.phar
 
-deps: download_models composer javascript_deps l10n_deps
+deps: download_models composer javascript_deps l10n_deps handlebars
```
stalker314314: Consider putting handlebars first, so it blows before it execute those "slow" stuff, like "download_models"

Makefile
```diff
@@ -26,6 +28,18 @@ else
...
 endif
 
+handlebars:
+ifeq (,$(handlebars))
+	@echo "No handlebars command available, remember to install this"
+	@echo " # npm install handlebars -g"
+	@echo "Then you must compile the templates"
+	@echo " # make js-templates"
+else
+	@echo "Remember compile the templates"
```
stalker314314: why we need to tell user to do this? Any reason why `js-templates` cannot be just another rule, something like this:
```
javascript_deps: js/handlebars.js js/lozad.js js/templates.js

js/templates.js:
	handlebars js/templates -f js/templates.js
```

Makefile
matiasdelellis: Well, Compiling the templates, must be a rule just as compile stylesheets from Sass/less.
Maybe we must add a 'build' rule, that download the dependencies, and build the templates, but these must be the rules separately.

Makefile
matiasdelellis: I put it at the end so that the message is seen. As it does not generate an error, it does not stop downloading the rest of the dependencies...

Makefile
matiasdelellis: The dependency is the handlebars binary. :sweat_smile: 
Let it be clear that this is only in development. Users from the appstore, nor should they know about this.

matiasdelellis:
I will modify it to check Handlebars as a dependency, and stop if it does not exist. and generate a 'build' rule that does everything necessary for the application to work...

Ok?

Makefile
stalker314314: makes sense, keep it:)

stalker314314:
"build" rule makes sense! BTW, I was under impression that current "deps" rule *is* "build" rule?:) But, OK.

BTW, do you plan to make "build" rule default one (instead of "deps" which is current default)? If it will **not** be default rule, consider updating documentation in README.md!

matiasdelellis:
Please @stalker314314 , look at it now.
 * Rename deps to build. It is the default rule
 * Separate the rules for translations, that are not needed unless you want to add a translation.
 * Move js deps and models to vendor folder.. Unnecessary here, but I wanted to do long ago. :sweat_smile: 
 * Add some comments to make it better understood.

PR #123, opened 2019-02-21
allow Nextcloud run in a subdirectory

feedthedogs:


js/personal.js
```diff
@@ -80,6 +80,7 @@ View.prototype = {
         var html = Handlebars.templates['personal']({
             loaded: this._persons.isLoaded(),
             persons: this._persons.getAll(),
+            appPath: OC.generateUrl('/'),
```
matiasdelellis: Hi @feedthedogs 
First of all, thank you very much for helping us !!. :smile: 

This was something that was pending .. :wink: 

Here you can generate two paths to avoid using '../../' :confused: in the template..

> appPath: OC.generateUrl('/apps/facerecognition') /* To thumbnails*/

.. and

> filePath: OC.generateUrl('/f') /* To open on file links */

js/templates/personal.handlebars
```diff
@@ -15,8 +15,8 @@
 			</h2>
 			<div class="persons-previews">
 				{{#each this.faces}}
-					<a target="_blank" href="/index.php/f/{{file-id}}">
-						<div class="face-preview" data-background-image="/index.php/apps/facerecognition/face/{{id}}/thumb/50" width="50" height="50"></div>
+					<a target="_blank" href="{{../../appPath}}f/{{file-id}}">
```
matiasdelellis: Here you can use `"{{filePath}}/{{file-id}}"`

js/templates/personal.handlebars
```diff
@@ -15,8 +15,8 @@
...
+						<div class="face-preview" data-background-image="{{../../appPath}}apps/facerecognition/face/{{id}}/thumb/50" width="50" height="50"></div>
```
matiasdelellis: And here: `"{{appPath}}/face/{{id}}/thumb/50"`

js/personal.js
feedthedogs: I tried what you suggested, but it didn't work.
The reason for the ../../ is the handlebar's variables are within the foreach sections
another option if you don't like that notation I think you can use root

matiasdelellis:
Sorry for the delay, tomorrow I confirm if this works... :wink:

matiasdelellis:
Hi @feedthedogs 
It should work better with the last branch merged.. you can check? Thanks again.!

PR #124, opened 2019-03-02
Use xenial in Travis & fix MySQL formatting

stalker314314:
* Added to use Xenial in Travis
* This led to uncovering one bug using tests - https://travis-ci.org/matiasdelellis/facerecognition/builds/500841309
* I had to fix one issue [I already suspected itmight be problem](https://github.com/matiasdelellis/facerecognition/blob/ddab4c5c865adabe8b67da47cac814cc4dfb52e9/lib/Db/Face.php#L177), but we somehow didn't hit it until now - it is different formatting for `DateTime` field type in MySQL vs. other DB types. Deck app already found way to fix it (as explained in comment), so I just used (very similar) their approach

### DIFF
--- a/lib/Db/Face.php
+++ b/lib/Db/Face.php
@@ -45,12 +45,6 @@ use OCP\AppFramework\Db\Entity;
 class Face extends Entity implements JsonSerializable {
 
 	/**
-	 * Database type, needed to format dates
-	 * @var string
-	 */
-	private $databaseType = 'sqlite';
-
-	/**
 	 * Image from this face originated from.
 	 *
 	 * @var int
@@ -126,10 +120,6 @@ class Face extends Entity implements JsonSerializable {
 		return $face;
 	}
 
-	public function setDatabaseType(string $type) {
-		$this->databaseType = $type;
-	}
-
 	/**
 	 * Helper method, to normalize face sizes back to original dimensions, based on ratio
 	 *
@@ -170,7 +160,7 @@ class Face extends Entity implements JsonSerializable {
 			'top' => $this->top,
 			'bottom' => $this->bottom,
 			'descriptor' => $this->descriptor,
-			'creation_time' => $this->getCreationTime(true)
+			'creation_time' => $this->creationTime
 		];
 	}
 
--- a/lib/Db/Face.php
+++ b/lib/Db/Face.php
@@ -45,6 +45,12 @@ use OCP\AppFramework\Db\Entity;
 class Face extends Entity implements JsonSerializable {
 
 	/**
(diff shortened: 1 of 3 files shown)


lib/Db/Face.php
```diff
@@ -173,9 +183,13 @@ public function setDescriptor($descriptor) {
...
 	}
 
-	public function getCreationTime(): string {
-		// Deck app have special handling for MySQL here:
+	public function getCreationTime(bool $isoFormat = false): string {
+		// Same special handling for MySQL as in Deck app
 		// https://github.com/nextcloud/deck/blob/139b38ca1df0ff17792eb218cc8ba7e3b04b4e51/lib/Db/Card.php#L84
+		//
+		if (!$isoFormat && $this->databaseType === 'mysql') {
+			return $this->creationTime->format('Y-m-d H:i:s');
```
matiasdelellis: Why not always convert to this full format?.
I'm looking at all the Nextcloud repositories, and only deck has this problem, and the rest, use a full formating.

stalker314314: Because we are 1) using DateTime and not plain timestamp, and 2) we use proper getter/setters that need to joggle between string and DateTimer.

I think, if we go with timestamps, that we might be able to completely make this redundant. If you want, I can try doing that, but that would be backward incompatible change (and not sure if we need upgrade step for something not released?). But, this is bigger change, I would need sign-off from you that you would accept that as a change

matiasdelellis: > I can try doing that, but that would be backward incompatible change (and not sure if we need upgrade step for something not released?)

That's exactly why I suggest it - the time to make incompatible changes is now, that we do not publish it yet.. :sweat_smile:

If you see it well, and you think it will not bring problems in the future, I approve it.

P.s: I know it will not bring problems, I would just prefer that it works the same with all the databases, and (being paranoid. haha.) I worry about the performance of the additional check. :sweat_smile:

matiasdelellis:
P.s 2
```
echo "deb [ trusted=yes ] http://repo.delellis.com.ar xenial main" > /etc/apt/sources.list.d/20-pdlib.list
apt update
apt install php7.0-pdlib
```

Not tested, due i not have any installation of xenial, but it should work .. :grimacing:

lib/Db/Face.php
stalker314314: So, I think I fixed "original" problem, take a look at https://github.com/matiasdelellis/facerecognition/pull/126. If we merge that, this one is not needed!

matiasdelellis: Ok..
Merge the branch #126 first, and then rebase it. :smile:

PR #130, opened 2019-03-18
Delegate translations to trasifex.. .

matiasdelellis:
:smile: 

https://www.transifex.com/matias/nextcloud-facerecognition


Makefile
```diff
@@ -129,6 +139,7 @@ test: build
 	./vendor/bin/phpunit --coverage-clover clover.xml -c phpunit.xml
 
 clean:
+	l10n-clean
```
stalker314314: just mind debugging this - are you sure that `make clean` will not call `l10-clean` command (instead of executign sub task?)

stalker314314:
awesome!:) How does this work - will transifex automatically push to master, or you/me need to do some manual actions?

Makefile
matiasdelellis: D'Oh!.
You are right. ..

matiasdelellis:
> awesome!:)

😬 

> How does this work - will transifex automatically push to master, or you/me need to do some manual actions?

It is not automatic, and I prefer to continue manual.. I hate follow some project, and to think that there were advances, but later I realize that the last 100 commits were automatic translations .. 😅 

l10n-transifex-pull:
  * Download translations of transifex. It will ask you the login information ..

l10n-transifex-apply:
 * convert the translation downloads into the js / json files used by nextcloud.. .. and it is the only thing that we must push to this repository,

l10n-update-pot: 
 * If we add new terms to translate we must update the source file (.pot) with this rule.

l10n-transifex-push:
 * Send the changes from the source to transifex to translate there.

Later I document it in the wiki.

matiasdelellis:
You are already the Serbian translator ..  ;)

PR #131, opened 2019-03-28
Implement settings controller and adds sensitivity settings to fronted..

matiasdelellis:
imagen

@stalker314314 
Now it just saves the value, but not force the regrouping. How do you imagine this?


matiasdelellis:
```
$recreateClusters = booltval($this->config->getAppValue('facerecognition', 'recreate-clusters', 'false'));
if ($haveStalePersons === false && $haveNewFaces === false && $recreateClusters === false) {
}
```
Here?
https://github.com/matiasdelellis/facerecognition/blob/more-settings/lib/BackgroundJob/Tasks/CreateClustersTask.php#L148

lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -98,6 +98,8 @@ public function execute(FaceRecognitionContext $context) {
 			$this->createClusterIfNeeded($user);
 		}
 
+		$this->config->setAppValue('facerecognition', 'recreate-clusters', 'false');
```
matiasdelellis: There is a problem ..
When it is executed for a single user, the recreation is disabled for all. So, this should should be a user settings? :confused:

matiasdelellis:
Ready for initial review.. 
From this branch, I will return to work on the option to enable by users .. :wink:

lib/Controller/SettingController.php
```diff
@@ -0,0 +1,102 @@
...
+				$this->userManager->callForSeenUsers(function(IUser $user) {
+					$this->config->setUserValue($user->getUID(), 'facerecognition', 'recreate-clusters', 'true');
+				});
+				break;
+			default:
+				break;
+		}
+
+		$result = [
+			'success' => 'true',
```
stalker314314: probably bikesheding at this point, but why is 'success' true if $value was not 'sensitivity' (if it fall through case switch).

P.S. we need something like S_FALSE in C :)

templates/settings/admin.php
```diff
@@ -0,0 +1,37 @@
...
+		<h2>
+			<?php p($l->t('Face Recognition'));?>
+			<a target="_blank" rel="noreferrer noopener" class="icon-info" title="<?php p($l->t('Open Documentation'));?>" href="https://github.com/matiasdelellis/facerecognition/wiki"></a>
+		</h2>
+		<div>
+			<h3>
+				<strong><?php p($l->t('Sensitivity'));?></strong>
+				<a target="_blank" rel="noreferrer noopener" class="icon-info" title="<?php p($l->t('Open Documentation'));?>" href="https://github.com/matiasdelellis/facerecognition/wiki/Sensitivity"></a>
+			</h3>
+			<p class="settings-hint"><?php p($l->t('The sensitivity determines how different the descriptors of the faces can be to consider them the same person.'));?></p>
```
stalker314314: #consider

`The sensitivity determines how different faces need to be to consider them as the same person. Lower values will be more precise and will not put different people under same person, but it will tend to create more groups, even for same person. Higher values can see different persons under single same person, but it will create less number of clusters.`

or 
`
The sensitivity determines how different faces need to be to consider them as the same person. See <a href="">wiki</a> for details`

matiasdelellis: I started writing a lot and then I preferred to leave it more concise by expanding on the wiki. The link is two lines of code above..

 * https://github.com/matiasdelellis/facerecognition/wiki/Sensitivity
 
> ...need to be..

The faces mo "need to be" different, at the most "can be" different.. 
.. but I agree to stop talking about the descriptors. :wink:

templates/settings/admin.php
```diff
@@ -0,0 +1,37 @@
...
+			<p>
+				<span><?php p($l->t('More sensible, more groups'));?></span>
```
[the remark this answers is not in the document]
matiasdelellis: We come using 'sensitivity', then I accept this. :wink: 
> More sensitivity, more groups,

lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -204,6 +211,8 @@ private function getCurrentClusters(array $faces): array {
 	private function getNewClusters(array $faces): array {
 		// Create edges for chinese whispers
 		$euclidean = new Euclidean();
+		$sensitivity = floatval($this->config->getAppValue('facerecognition', 'sensitivity', '0.5'));
```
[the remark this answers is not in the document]
matiasdelellis: Just because it is in the middle of the scale .. :sweat_smile: 

0.4 <= sensitivity <= 0.6 

I guess it does not make sense to go beyond those values. . No?
Then it will be the administrator's choice ..

lib/Controller/SettingController.php
matiasdelellis: > probably bikesheding at this point, but why is 'success' true if $value was not 'sensitivity' (if it fall through case switch).

I suppose that as we add preferences we can improve this.

> P.S. we need something like S_FALSE in C :)

Maybe I can pass the Boolean directly. I do not remember why I used string here.

matiasdelellis: Ahh..
> Why is 'success' true if $value was not 'sensitivity'

The setting always changes, you can send garbage that changes it, and that's why it returns ok .. the switch comes to consider the special cases, when you have to do something else when changing a particular option. In this case, recreate the clusters

matiasdelellis:
@stalker314314 
Thanks for review.. See again..  :smiley:

PR #133, opened 2019-03-29
Second try..  Add check box to enable analysis by user..

matiasdelellis:
Ready to first review, but need help.. 

The only difficult case is when the user deactivates the analysis while it is running.. How do we handle it? 'If ($userEnabled)? everywhere?

### DIFF
--- a/js/personal.js
+++ b/js/personal.js
@@ -119,8 +119,8 @@ View.prototype = {
                 if (enabled) {
                     OC.Notification.showTemporary(t('facerecognition', 'The analysis is enabled, please be patient, you will soon see your friends here.'));
                 } else {
-                    OC.Notification.showTemporary(t('facerecognition', 'The analysis is disabled, we eliminate all information for the recognition of your friends.'));
-                    self.reload();
+                    OC.Notification.showTemporary(t('facerecognition', 'The analysis is disabled. In the next execution your data will be completely eliminated.'));
+                    location.reload();
                 }
             }
         });
--- a/js/personal.js
+++ b/js/personal.js
@@ -134,13 +134,18 @@ View.prototype = {
             welcomeHint: t('facerecognition', 'Here you can see photos of your friends that are recognized'),
             enableDescription: t('facerecognition', 'Analyze my images and group my loved ones with similar faces'),
             loadingMsg: t('facerecognition', 'Looking for your recognized friends'),
-            loadingIcon: OC.imagePath('core', 'loading.gif'),
-            emptyMsg: t('facerecognition', 'Your friends have not been recognized yet'),
-            emptyHint: t('facerecognition', 'Please, be patient')
+            loadingIcon: OC.imagePath('core', 'loading.gif')
         };
 
-        if (this._persons.isEnabled() === 'true')
+        if (this._persons.isEnabled() === 'true') {
             context.enabled = true;
+            context.emptyMsg = t('facerecognition', 'Your friends have not been recognized yet');
+            context.emptyHint = t('facerecognition', 'Please, be patient');
+        }
+        else {
+            context.emptyMsg = t('facerecognition', 'The analysis is disabled');
+            context.emptyHint = t('facerecognition', 'Enable it to find your loved ones');
+        }
 
         if (this._persons.getActive() !== undefined)
             context.person = this._persons.getActive();
--- a/js/personal.js
+++ b/js/personal.js
@@ -3,6 +3,13 @@
 
 $(document).ready(function () {
 
+const state = {
+    OK: 0,
+    FALSE: 1,
+    SUCCESS: 2,
+    ERROR:  3
+}
+
 /*
  * Faces in memory handlers.
  */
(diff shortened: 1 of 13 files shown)


js/personal.js
```diff
@@ -94,19 +106,42 @@ View.prototype = {
...
+            url: OC.generateUrl('apps/facerecognition/setuservalue'),
+            data: {
+                'type': 'enabled',
+                'value': enabled
+            },
+            success: function () {
+                if (enabled) {
+                    OC.Notification.showTemporary(t('facerecognition', 'The analysis is enabled, please be patient, you will soon see your friends here.'));
+                } else {
+                    OC.Notification.showTemporary(t('facerecognition', 'The analysis is disabled, we eliminate all information for the recognition of your friends.'));
```
stalker314314: eliminate -> eliminated? I would even go with "removed all face data"

stalker314314:
> The only difficult case is when the user deactivates the analysis while it is running.. How do we handle it? 'If ($userEnabled)? everywhere?

so, if we are in the middle of the task, who know what can happen, probably some crash due to foreign keys:) We can either delay deletion after task is done (you don't want to do this, if I understand you, you want *immediate* action) or we can add another task, after all other tasks are done, that will check, for all users, if user is disabled and do reset again (just in case there is some leftovers).

js/personal.js
matiasdelellis: I think you already suggested it in another branch.. :sweat_smile: 
Ok.

matiasdelellis:
> so, if we are in the middle of the task, who know what can happen, probably some crash due to foreign keys:) 

Exactly .. I have no idea .. but it can not be good .. :see_no_evil: 

> We can either delay deletion after task is done (you don't want to do this, if I understand you, you want immediate action) or we can add another task, after all other tasks are done, that will check, for all users, if user is disabled and do reset again (just in case there is some leftovers).

I guess for a first version, we can delete the scheduled task ..
And in the notification of the fronted, say something like this:
 * _They will be completely eliminated in the next analysis, after that, if you re-enable it, you will start from scratch._

lib/Controller/SettingController.php
```diff
@@ -69,6 +70,8 @@ public function setUserValue($type, $value) {
 			case 'enabled':
 				if ($value === 'false')
 					$this->faceManagementService->resetAllForUser($this->userId);
+				else
```
stalker314314: nice catch!:)

matiasdelellis:
Hi @stalker314314 
I still have little time, but at least I will try to finish this. 

About the backend, I think it's all necessary and it's fine.. but please review it again.

matiasdelellis:
Well,
Still is necessary to correct the tests, since as default is disabled, it never works ...

js/personal.js
```diff
@@ -94,19 +106,42 @@ View.prototype = {
...
+            url: OC.generateUrl('apps/facerecognition/setuservalue'),
+            data: {
+                'type': 'enabled',
+                'value': enabled
+            },
+            success: function () {
+                if (enabled) {
+                    OC.Notification.showTemporary(t('facerecognition', 'The analysis is enabled, please be patient, you will soon see your friends here.'));
+                } else {
+                    OC.Notification.showTemporary(t('facerecognition', 'The analysis is disabled. In the next execution your data will be completely eliminated.'));
```
stalker314314: consider rewording `your data will be completely eliminated` to something less scary, for example `The analysis is disabled. We will remove data about all your recognized faces.` (this way, it seems like some user data might be removed, and "eliminated" seems like something that Daleks from Dr. Who would say:D)

stalker314314: or, as you said in other place, "You will lose all the information analyzed"

lib/Controller/PersonController.php
```diff
@@ -63,6 +63,8 @@ public function index() {
 		$model = intval($this->config->getAppValue('facerecognition', 'model', AddDefaultFaceModel::DEFAULT_FACE_MODEL_ID));
 
 		$resp = array();
+		$resp['enabled'] = $this->config->getUserValue($this->userId, 'facerecognition', 'enabled', false);
```
stalker314314: if not enabled, should we send persons to client at all? Why just not "fake" it and send empty array?

lib/BackgroundJob/Tasks/DisabledUserRemovalTask.php
```diff
@@ -0,0 +1,105 @@
...
+use OCP\IUser;
+
+use OCA\FaceRecognition\BackgroundJob\FaceRecognitionBackgroundTask;
+use OCA\FaceRecognition\BackgroundJob\FaceRecognitionContext;
+use OCA\FaceRecognition\Db\ImageMapper;
+use OCA\FaceRecognition\Migration\AddDefaultFaceModel;
+
+use OCA\FaceRecognition\FaceManagementService;
+
+/**
```
stalker314314: is comment stale?

lib/Controller/PersonController.php
matiasdelellis: Oh .. You're right... :+1:

lib/BackgroundJob/Tasks/DisabledUserRemovalTask.php
matiasdelellis: Ohh.. Thanks. :disappointed:

matiasdelellis:
Please,
Last review and merge it.. :smiley: 

With the latest commits, improve the user experience.. . and we can continue with other things.

lib/BackgroundJob/Tasks/DisabledUserRemovalTask.php
```diff
@@ -0,0 +1,105 @@
...
+				$eligable_users[] = $user->getUID();
+			});
+		} else {
+			$eligable_users[] = $this->context->user->getUID();
+		}
+
+		// Reset user datas if needed.
+		foreach($eligable_users as $userId) {
+			$userEnabled = $this->config->getUserValue($userId, 'facerecognition', 'enabled', 'false');
+			$imageCount = $this->imageMapper->countUserImages($userId, $model);
```
stalker314314: Ideally, you want to check here for personCount too (persons are top-level entity and can exist without any image or face). I mean, normally, if they exist and there is no images, it's a bug, but...if there is a bug, no way user can delete those persons.

matiasdelellis: Let's reproduce the error before adding ugly unnecessary fixes.

PR #135, opened 2019-03-30
Enable pdlib in travis and image processing tests

stalker314314:
As @matiasdelellis already did massive exploration to enable pdlib in travis, I am just copy/pasting his work.

Also added image processing tests (now that we can have them!) + simple requirement test.

These tests did uncover some product issues, so those are also fixed here. One is bogus naming of variables in Image entity. Other is that mysql/sqlite return string ('1') for booleans (I think this is something with doctrine more than with mysql, as boolean is tinybit there). I didn't get to the bottom of it, but my proposed solution looks sane in any case, so I am calling it a day:)


stalker314314:
I will create clean one

PR #136, opened 2019-03-30
Enable pdlib in travis and add image processing tests, closes #86

### LINKED ISSUE #86: PDlib debian packages
Maybe I should add this issue in PDLib repository, but I find it much better to add it here, ha ha. Assigned to you @matiasdelellis (feel free to remove assignee label:D). I think this is blocker for beta version, as this can help with installations. This can also help to be written in FAQ and it will unblock bunch of integration tests.

BTW, where will those debian packages be hosted? LMK if you need some VMs or any other help for this task!

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67911486-pdlib-debian-packages?utm_campaign=plugin&utm_content=tracker%2F74944432&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F74944432&utm_medium=issues&utm_source=github).
</bountysource-plugin>

stalker314314:
As @matiasdelellis already did massive exploration to enable pdlib in travis, I am just copy/pasting his work.

Also added image processing tests (now that we can have them!) + simple requirement test.

These tests did uncover some product issues, so those are also fixed here. One is bogus naming of variables in Image entity. Other is that mysql/sqlite return string ('1') for booleans (I think this is something with doctrine more than with mysql, as boolean is tinybit there). I didn't get to the bottom of it, but my proposed solution looks sane in any case, so I am calling it a day:)


stalker314314:
> If it pass in travis correctly .. Great..

To be honest, I was also surprised how good it works:D (knock, knock)

stalker314314:
BTW, our code coverage got +10% boost:D

Issue #137, opened 2019-04-05
I have orphan persons (without faces), and always recreates the group.


matiasdelellis:
## Graphic description
![imagen](https://user-images.githubusercontent.com/733715/55597490-3ddf6300-5724-11e9-97dd-f5f7021bd431.png)

Just that.. :disappointed: 

**Query to test:**

```
SELECT * FROM `oc_face_recognition_persons` P
WHERE NOT EXISTS(SELECT * FROM `oc_face_recognition_faces` F
                 WHERE F.person = P.id)
```

**Response here:**

id | user | name | is_valid | last_generation_time | linked_user
-- | -- | -- | -- | -- | --
4736 | matias | Vicky | 0 | 2019-03-29 16:59:53 | NULL
4737 | matias | Sofy | 0 | 2019-03-29 16:59:53 | NULL
5114 | matias | Matias De lellis | 0 | 2019-04-01 16:48:39 | NULL

Surely it is due to so many tests of the change of the sensitivity parameter, but I still can not reproduce to document it correctly

matiasdelellis:
I still do not see how they were orphans, but now I understand why they are never erased.

The current clusters are formed from the faces. (Without taking into account the persons table):

https://github.com/matiasdelellis/facerecognition/blob/master/lib/BackgroundJob/Tasks/CreateClustersTask.php#L203

As you identify which one should be removed from this list, you will never be able to eliminate the orphans:
https://github.com/matiasdelellis/facerecognition/blob/master/lib/Db/PersonMapper.php#L166

@stalker314314 
Any ideas?

stalker314314:
Hard problem:) To be honest, I have no idea how this can happen. Couple of things:
* What you suggested (with getCurrentClusters), I think, is not true, as there should be no orphan persons to begin with! So, if there was orphan person, they will reappear. If there was no orphans, `getCurrentClusters` should return no orphans. THis might keep error going, but it was not root cause.
* I don't know where problem is, but I _suspect_ it to be in `mergeClusterToDatabase` (I already fixed nasty bug there https://github.com/matiasdelellis/facerecognition/commit/ff2ddc4db12d6fa7949173e6962dd52ead18d0ac and it is completely untested, in contrast to rest of `CreateClustersTask`)
[...]

PR #138, opened 2019-04-06
Adding more tests for create clusters

stalker314314:
In the hope it can reveal something for #137 . It didn't:(, but I want to add those tests anyway, they should be nice addition.

### DIFF
--- a/lib/BackgroundJob/Tasks/CreateClustersTask.php
+++ b/lib/BackgroundJob/Tasks/CreateClustersTask.php
@@ -96,6 +96,7 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 
 		foreach($eligable_users as $user) {
 			$this->createClusterIfNeeded($user);
+			yield;
 		}
 
 		return true;
@@ -159,6 +160,10 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 				$this->logInfo('Clusters already exist, but there was some change that requires recreating the clusters');
 			}
 		} else {
+			// User should not be able to use this directly, used in tests
+			$forceCreateClusters = $this->config->getUserValue($userId, 'facerecognition', 'force-create-clusters', 'false');
+			$forceCreation = ($forceCreateClusters === 'true');
+
 			// These are basic criteria without which we should not even consider creating clusters.
 			// These clusters will be small and not "stable" enough and we should better wait for more images to come.
 			// todo: 2 queries to get these 2 counts, can we do this smarter?
@@ -170,7 +175,7 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 			}
 			$facesCount = $this->faceMapper->countFaces($userId, $modelId);
 			// todo: get rid of magic numbers (move to config)
-			if (($facesCount < 1000) && ($imageCount < 100) && ($percentImagesProcessed < 0.95)) {
+			if (!$forceCreation && ($facesCount < 1000) && ($imageCount < 100) && ($percentImagesProcessed < 0.95)) {
 				$this->logInfo(
 					'Skipping cluster creation, not enough data (yet) collected. ' .
 					'For cluster creation, you need either one of the following:');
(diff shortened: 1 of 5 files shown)


lib/Db/FaceMapper.php
```diff
@@ -197,5 +188,6 @@ public function insertFace(Face $face, IDBConnection $db = null) {
 				'creation_time' => $qb->createNamedParameter($face->creationTime, IQueryBuilder::PARAM_DATE),
 			])
 			->execute();
+		$face->setId((int) $qb->getLastInsertId());
```
matiasdelellis: Why is this?
It is completely logical, but why not add it before?

lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -159,6 +160,10 @@ private function createClusterIfNeeded(string $userId) {
 				$this->logInfo('Clusters already exist, but there was some change that requires recreating the clusters');
 			}
 		} else {
+			// User should not be able to use this directly, used in tests
+			$forceCreateClusters = $this->config->getUserValue($userId, 'facerecognition', 'force-create-clusters', 'false');
+			$forceCreation = ($forceCreateClusters === 'true');
```
matiasdelellis: I do not like to add this, but I guess it's acceptable ..

tests/integration/CreateClustersTaskTest.php
```diff
@@ -0,0 +1,147 @@
...
+use OCA\FaceRecognition\BackgroundJob\FaceRecognitionContext;
+use OCA\FaceRecognition\BackgroundJob\FaceRecognitionLogger;
+use OCA\FaceRecognition\BackgroundJob\Tasks\CreateClustersTask;
+use OCA\FaceRecognition\Db\Face;
+use OCA\FaceRecognition\Db\Image;
+use OCA\FaceRecognition\Migration\AddDefaultFaceModel;
+
+use Test\TestCase;
+
+class CreateClustersTaskTest extends TestCase {
```
matiasdelellis: Great!.

tests/integration/MergeClusterToDatabaseTest.php
```diff
@@ -0,0 +1,695 @@
...
+use OCA\FaceRecognition\BackgroundJob\FaceRecognitionLogger;
+use OCA\FaceRecognition\BackgroundJob\Tasks\CreateClustersTask;
+use OCA\FaceRecognition\Db\Face;
+use OCA\FaceRecognition\Db\Image;
+use OCA\FaceRecognition\Db\Person;
+use OCA\FaceRecognition\Migration\AddDefaultFaceModel;
+
+use Test\TestCase;
+
+class MergeClusterToDatabaseTest extends TestCase {
```
matiasdelellis: Geat..
Pretty complete .. :+1:

lib/Db/FaceMapper.php
stalker314314: We didn't added it before (when we wrote this), because, I guess, we never needed id of inserted face before:D

tests/integration/MergeClusterToDatabaseTest.php
stalker314314: +9% in tests too:D

lib/BackgroundJob/Tasks/CreateClustersTask.php
stalker314314: yeah, it is screaming "this code is added because of tests":) Now, that I am doing stress tests, I will read more about mocking capabilities of PHPUnit, so I might cook something with mocks (instead of configs like this).

stalker314314:
waiting for @matiasdelellis signoff, to check-in:)

lib/Db/FaceMapper.php
matiasdelellis: Ok ok..
:sweat_smile:

lib/BackgroundJob/Tasks/CreateClustersTask.php
matiasdelellis: Wow.. Ok..

PR #142, opened 2019-05-09
Changing logic for landmark detection to work on uncropped images (fixes #140, #20)

### LINKED ISSUE #140: Crop the faces for analysis changes the result of the descriptors
Hi @stalker314314 

I had this test quite ago but now polishes it a bit ..

**To investigate:**
I simply reproduced the crop and analysis code, and compared it with not cropping the the same image.

The interesting thing, is both methods have a distance of **0.11836271306919** ¿¿¡!??? :confused: 
And we are discussing between using 0.5 or 0.4 sensitivity.. :sweat_smile: 

### Test:
```
[matias@nube facerecognition]# sudo -u apache php test.php 
We will calculate the descriptors in the same way that NC Facerecognition does... Done.
Now we will calculate the descriptors with default pdlib method... Done.
The distance between both methods is: 0.11836271306919
```

### Source code:
<details><summary>Test Source code:</summary>
<p>

```
<?php
$img_path = "linus.jpeg";
$img_crop_path = "cropped_" . $img_path ;

$cfd = new CnnFaceDetection("vendor/models/1/mmod_human_face_detector.dat");

// Find faces.
// Source: https://github.com/matiasdelellis/facerecognition/blob/master/lib/BackgroundJob/Tasks/ImageProcessingTask.php#L222
$facesFound = $cfd->detect($img_path);
foreach($facesFound as $faceFound)
{
    print ("We will calculate the descriptors in the same way that NC Facerecognition does... ");

    // Get rect to crop image.
    // Source: https://github.com/matiasdelellis/facerecognition/blob/c41bf9c46bd479c28dea8d751c5923d18b0493a9/lib/Db/Face.php#L114
    $left   = max($faceFound["left"], 0);
    $right  = $faceFound["right"];
    $top    = max($faceFound["top"], 0);
    $bottom = $faceFound["bottom"];

    $width  = $right - $left;
    $height = $bottom - $top;

    // Crop image.
    // Source: https://github.com/matiasdelellis/facerecognition/blob/master/lib/BackgroundJob/Tasks/ImageProcessingTask.php#L335
    $image = imagecreatefromjpeg($img_path);
    $imageCrop = imagecrop($image, [
        'x'      => (int)$left,
        'y'      => (int)$top,
        'width'  => (int)$width,
        'height' => (int)$height
    ]);
    imagejpeg($imageCrop, $img_crop_path, 100);

    // Get landmarks amd compute descriptors
    // Source: https://github.com/matiasdelellis/facerecognition/blob/master/lib/BackgroundJob/Tasks/ImageProcessingTask.php#L323
    $fld = new FaceLandmarkDetection("vendor/models/1/shape_predictor_5_face_landmarks.dat");
    $fr = new FaceRecognition("vendor/models/1/dlib_face_recognition_resnet_model_v1.dat");

    $landmarks = $fld->detect($img_crop_path, array(
        "left"   => 0,
        "top"    => 0,
        "bottom" => (int)$height,
        "right"  => (int)$width
    ));
    $descriptor = $fr->computeDescriptor($img_crop_path, $landmarks);

    print ("Done.\n");
    print ("Now we will calculate the descriptors with default pdlib method... ");

    // pdlib method.
    // Source: https://github.com/goodspb/pdlib#general-usage
    $fld = new FaceLandmarkDetection("vendor/models/1/shape_predictor_5_face_landmarks.dat");
    $full_landmarks = $fld->detect($img_path, $faceFound);
    $descriptorAligned = $fr->computeDescriptor($img_path, $full_landmarks);

    print ("Done.\n");

    // Calculate distance.
    // Source: https://github.com/matiasdelellis/facerecognition/blob/master/lib/Helper/Euclidean.php#L32
    $n = count($descriptor);
    $sum = 0;
    for ($i = 0; $i < $n; $i++) {
        $sum += ($descriptorAligned[$i] - $descriptor[$i]) * ($descriptorAligned[$i] - $descriptor[$i]);
    }

    print("The distance between both methods is: " . sqrt($sum) . "\n");
}
```

</p>
</details>

### Test image:
Image tested

### Croped Face
Cropped face with NC method:

#### Resume
I do not say that one is better than another, only that they are different, and I check this with many photos.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/72972007-crop-the-faces-for-analysis-changes-the-result-of-the-descriptors?utm_campaign=plugin&utm_content=tracker%2F74944432&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F74944432&utm_medium=issues&utm_source=github).
</bountysource-plugin>

stalker314314:
This should fix problem with cropping that you detected, @matiasdelellis!
As discussed in #140, landmarks and descriptors were pulled from cropped version of image. There is no need for that. This should fix:
* unnecessary saving to temp (second time)
* more precise landmarks and descriptors as we don't crop anything
* hopefully better clustering

For 2 and 3, I am not sure, as I don't know if dlib actually uses points in image outside of face boundaries, but my testing shows that generated clusters were not degraded, I mean, they were as good as they were previously (although hard to quantify how "better" they are)

### DIFF
--- a/lib/BackgroundJob/Tasks/ImageProcessingTask.php
+++ b/lib/BackgroundJob/Tasks/ImageProcessingTask.php
@@ -300,7 +300,7 @@ class ImageProcessingTask extends FaceRecognitionBackgroundTask {
 		}
 
 		$this->logDebug(sprintf('Image scaled from %dx%d to %dx%d (since max image area is %d pixels^2)',
-			$newWidth, $newHeight, $widthOrig, $heightOrig, $maxImageArea));
+			$widthOrig, $heightOrig, $newWidth, $newHeight, $maxImageArea));
 
 		return 1 / $scaleFactor;
 	}
@@ -316,30 +316,22 @@ class ImageProcessingTask extends FaceRecognitionBackgroundTask {
 		$faces = $imageProcessingContext->getFaces();
 
 		foreach($faces as &$face) {
-			$tempfilePath = $this->cropFace($imageProcessingContext->getImagePath(), $face);
+			// For each face, we want to detect landmarks and compute descriptors.
+			// We use already resized image (from temp, used to detect faces) for this.
+			// (better would be to work with original image, but that will require
+			// another orientation fix and another save to the temp)
+			// But, since our face coordinates are already changed to align to original image,
+			// we need to fix them up to align them to temp image here.
+			$normalizedFace = clone $face;
+			$normalizedFace->normalizeSize(1.0 / $imageProcessingContext->getRatio());
 
-			// Usually, second argument to detect should be just $face. However, since we are doing image acrobatics
-			// and already have cropped image, bounding box for landmark detection is now complete (cropped) image!
-			$landmarks = $fld->detect($tempfilePath, array(
-				"left" => 0, "top" => 0, "bottom" => $face->height(), "right" => $face->width()));
-			$descriptor = $fr->computeDescriptor($tempfilePath, $landmarks);
+			// We are getting face landmarks from already prepared (temp) image (resized and with orienation fixed).
+			$landmarks = $fld->detect($imageProcessingContext->getTempPath(), array(
+				"left" => $normalizedFace->left, "top" => $normalizedFace->top,
+				"bottom" => $normalizedFace->bottom, "right" => $normalizedFace->right));
+			$descriptor = $fr->computeDescriptor($imageProcessingContext->getTempPath(), $landmarks);
 			$face->descriptor = $descriptor;
 		}
-	}
-
-	private function cropFace(string $imagePath, Face $face): string {
-		// todo: we are loading same image two times, fix this
-		$image = new OCP_Image(null, $this->context->logger->getLogger(), $this->context->config);
-		$image->loadFromFile($imagePath);
-		$image->fixOrientation();
-		$success = $image->crop($face->left, $face->top, $face->width(), $face->height());
-		if ($success === false) {
-			throw new \RuntimeException("Error during image cropping");
-		}
-
-		$tempfile = $this->tempManager->getTemporaryFile(pathinfo($imagePath, PATHINFO_EXTENSION));
-		$image->save($tempfile);
-		return $tempfile;
 	}
 
 	/**


stalker314314:
Ha ha, just figured it out, this commit will also close #20 :)

matiasdelellis:
Hi @stalker314314 
Thank you for taking this task.. :smile: 

I had done something similar, but did not have time to test. :disappointed: 
The only difference is that I change prepareImage() to not increase the size of the images when they are smaller that maxImageArea..

```
	private function prepareImage(string $imagePath) {
		$image = new OCP_Image(null, $this->context->logger->getLogger(), $this->context->config);
		$image->loadFromFile($imagePath);

		// Validate image
		$image->fixOrientation();
		if (!$image->valid()) {
			throw new \RuntimeException("Image is not valid, probably cannot be loaded");
		}

		$widthOrig = imagesx($image->resource());
		$heightOrig = imagesy($image->resource());
		if (($widthOrig <= 0) || ($heightOrig <= 0)) {
			$message = "Image is having non-positive width or height, cannot continue";
			$this->logInfo($message);
			throw new \RuntimeException($message);
		}

		// Ignore processing of images that are not large enough.
		$minImageSize = intval($this->config->getAppValue('facerecognition', 'min_image_size', '512'));
		if (($widthOrig < $minImageSize) || ($heightOrig < $minImageSize)) {
			return new ImageProcessingContext($imagePath, "", -1, true);
		}

		// Use the original image if it meets the memory parameters.
		$maxImageArea = $this->getMaxImageArea();
		if ($widthOrig*$heightOrig < $maxImageArea)
			return new ImageProcessingContext($imagePath, $imagePath, 1.0, false);

		// In any case, resize it to reduce memory consumption.
		$ratio = $this->resizeImage($image, $maxImageArea);
		$tempfile = $this->tempManager->getTemporaryFile(pathinfo($imagePath, PATHINFO_EXTENSION));
		$image->save($tempfile);
		return new ImageProcessingContext($imagePath, $tempfile, $ratio, false);
	}
```

lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -300,7 +300,7 @@ public function resizeImage(OCP_Image $image, int $maxImageArea): float {
 		}
 
 		$this->logDebug(sprintf('Image scaled from %dx%d to %dx%d (since max image area is %d pixels^2)',
-			$newWidth, $newHeight, $widthOrig, $heightOrig, $maxImageArea));
+			$widthOrig, $heightOrig, $newWidth, $newHeight, $maxImageArea));
```
matiasdelellis: Wow.. :sweat_smile:

lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -316,32 +316,24 @@ private function populateDescriptors(\FaceLandmarkDetection $fld, \FaceRecogniti
...
-			$landmarks = $fld->detect($tempfilePath, array(
-				"left" => 0, "top" => 0, "bottom" => $face->height(), "right" => $face->width()));
-			$descriptor = $fr->computeDescriptor($tempfilePath, $landmarks);
+			// For each face, we want to detect landmarks and compute descriptors.
+			// We use already resized image (from temp, used to detect faces) for this.
+			// (better would be to work with original image, but that will require
+			// another orientation fix and another save to the temp)
+			// But, since our face coordinates are already changed to align to original image,
+			// we need to fix them up to align them to temp image here.
+			$normalizedFace = clone $face;
```
matiasdelellis: You can use the original without cloning, since it is not used again..
But in any case, can not we normalize the faces to the original image directly when we detect them and work with the normalized ones?

matiasdelellis: Here we do it from the beginning, but should not the ratio be the reverse? Now :confused: 

https://github.com/matiasdelellis/facerecognition/blob/496da698288c7bb2666f36278177c7e01fe8fa36/lib/BackgroundJob/Tasks/ImageProcessingTask.php#L228

stalker314314: > You can use the original without cloning

Can I? I was not thinking I could (although my PHP is not good). I am changing fields of `$face` object in-place (in `normalizeSize`) and we use those fields later when we are inserting face in DB (at the end of flow). So, by calling normalize on existing object, values would be changing and we would insert values from resized image, not from original image. Of course, I am talking this out of my ass now:), but here is sample:
https://ideone.com/KPWFs7

> Here we do it from the beginning

This is happening when faces are detected, and before landmark detection. FindFaces is nicely encapsulated to return list of faces with *proper* coordinates (as in original image), so it is hiding (by design) details of resized image on which we are doing face detection. I think it should stay that way. And in method I changed now, I am reversing that ratio that we changed in FindFaces to get us back to resized image coordinates. Not sure if this is what you are asking?

PR #144, opened 2019-05-29
Add memory limit settings to admin page.

matiasdelellis:
Thinking a little, I guess that the administrator's setting should be in terms of memory -which is the real limitation-, and not in terms of the redimensioning of the image. However I want to leave the maxImageArea as an advanced settings.

TODO:
 - [x] It is still ignored by the background task.
 - [x] Validate in setting controler.

It should already be usable.. You agree?.

### DIFF
--- a/lib/BackgroundJob/Tasks/CheckRequirementsTask.php
+++ b/lib/BackgroundJob/Tasks/CheckRequirementsTask.php
@@ -78,55 +78,53 @@ class CheckRequirementsTask extends FaceRecognitionBackgroundTask {
 		}
 
 		// Determine the system memory with some considerations.
-		$memory = MemoryLimits::getAvailableMemory();
-		if ($memory <= 0) {
+		$memoryAvailable = MemoryLimits::getAvailableMemory();
+		if ($memoryAvailable <= 0) {
 			// We cannot determine amount of memory to give to face recognition CNN.
 			// We will hardcode it here to 1GB, but plan is to expose this to user in future.
 			// TODO: allow user to choose "recognition quality", which will map to given memory.
 			// If user explicitely set something, we ignore getting memory from system.
 			$this->logDebug("Cannot detect amount of memory given to PHP process. Using 1GB for image processing");
-			$memory = 1024 * 1024 * 1024;
+			$memoryAvailable = 1024 * 1024 * 1024;
+		}
 
 		// Apply some prudent limits.
-		if ($memory < 1024 * 1024 * 1024) {
+		if ($memoryAvailable < 1024 * 1024 * 1024) {
 			$error_message =
 				"\n" .
-				"Seems that you have only " . intval($memory / (1024 * 1024)). "MB of memory given to PHP.\n" .
+				"Seems that you have only " . intval($memoryAvailable / (1024 * 1024)). "MB of memory given to PHP.\n" .
 				"Face recognition application requires at least 1 GB. You need to change your memory_limit in php.ini.\n" .
 				"Check https://secure.php.net/manual/en/ini.core.php#ini.memory-limit for details.\n" .
 				"If you already set this to unlimited, it seems your system is not having enough RAM memory.";
 			$this->logInfo($error_message);
 			return false;
 		} else {
-			$this->logDebug(sprintf('Found %d MB available to PHP.', $memory / (1024 * 1024)));
+			$this->logDebug(sprintf('Found %d MB available to PHP.', $systemMemory / (1024 * 1024)));
 			// No point going above 4GB ("4GB should be enough for everyone")
-			if ($memory > 4 * 1024 * 1024 * 1024) {
+			if ($memoryAvailable > 4 * 1024 * 1024 * 1024) {
 				$this->logDebug('Capping used memory to maximum of 4GB');
-				$memory = 4 * 1024 * 1024 * 1024;
+				$memoryAvailable = 4 * 1024 * 1024 * 1024;
 			}
 		}
 
 		// Apply admin setting limit.
 		$memorySetting = $this->config->getAppValue('facerecognition', 'memory-limits', '-1');
 		if ($memorySetting > 0) {
-			if ($memorySetting > $memory) {
+			if ($memorySetting > $memoryAvailable) {
 				$error_message =
 					"\n" .
-					"Seems that you configured" . intval($memorySetting / (1024 * 1024)) . " MB of memory, however, this exede"
-					"detected as maximum for PHP: " . intval($memory / (1024 * 1024)). "MB.\n" .
-					"You need to change your memory_limit in php.ini or 'Memory Limits' in settings.\n" .
-					"Check https://secure.php.net/manual/en/ini.core.php#ini.memory-limit for details.\n" .
-					"If you already set this to unlimited, it seems your system is not having enough RAM memory.";
+					"Seems that you configured" . intval($memorySetting / (1024 * 1024)) . " MB of memory, however, this exceeds"
+					"detected as maximum for PHP: " . intval($memoryAvailable / (1024 * 1024)). "MB.\n" .
+					"We will ignore the FaceRecognition settings, and we will limit to that.";
 				$this->logInfo($error_message);
-				return false;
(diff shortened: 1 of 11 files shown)


lib/BackgroundJob/Tasks/CheckRequirementsTask.php
```diff
@@ -105,8 +97,37 @@ public function execute(FaceRecognitionContext $context) {
...
+			}
 		}
 
+		// Apply admin setting limit.
+		$memorySetting = $this->config->getAppValue('facerecognition', 'memory-limits', '-1');
+		if ($memorySetting > 0) {
+			if ($memorySetting > $memory) {
+				$error_message =
+					"\n" .
+					"Seems that you configured" . intval($memorySetting / (1024 * 1024)) . " MB of memory, however, this exede"
```
stalker314314: exede <=> exceeds

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
```diff
@@ -77,6 +77,7 @@ public function execute(FaceRecognitionContext $context) {
 			return false;
 		}
 
+		// Determine the system memory with some considerations.
 		$memory = MemoryLimits::getAvailableMemory();
```
stalker314314: now that we have `$memory` and `$memorySetting`, consider naming them differently. For example, `$systemMemoryLimit` and `$userMemoryLimit`?

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
```diff
@@ -105,8 +97,37 @@ public function execute(FaceRecognitionContext $context) {
...
+		$memorySetting = $this->config->getAppValue('facerecognition', 'memory-limits', '-1');
+		if ($memorySetting > 0) {
+			if ($memorySetting > $memory) {
+				$error_message =
+					"\n" .
+					"Seems that you configured" . intval($memorySetting / (1024 * 1024)) . " MB of memory, however, this exede"
+					"detected as maximum for PHP: " . intval($memory / (1024 * 1024)). "MB.\n" .
+					"You need to change your memory_limit in php.ini or 'Memory Limits' in settings.\n" .
+					"Check https://secure.php.net/manual/en/ini.core.php#ini.memory-limit for details.\n" .
+					"If you already set this to unlimited, it seems your system is not having enough RAM memory.";
```
stalker314314: > If you already set this to unlimited

As per this code, this cannot happen (this is under `($memorySetting > 0)`, right?

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
```diff
@@ -105,8 +97,37 @@ public function execute(FaceRecognitionContext $context) {
...
+				$this->logDebug('Capping used memory to maximum of 4GB');
+				$memory = 4 * 1024 * 1024 * 1024;
+			}
 		}
 
+		// Apply admin setting limit.
+		$memorySetting = $this->config->getAppValue('facerecognition', 'memory-limits', '-1');
+		if ($memorySetting > 0) {
+			if ($memorySetting > $memory) {
+				$error_message =
```
stalker314314: I don't know about this error. I mean, if user have 2GB and configured it to 4GB, I would use 2GB and just output some warning that user limit is greater than system limit. Intent of user limit should be not to overload system above some value, but if system/PHP cap is below that, I don't think of it as a problem. In other words, limit should be `min(userLimit, memoryLimit)`. 

So, I would just output warning here, but would not block execution or error out

templates/settings/admin.php
```diff
@@ -25,6 +25,20 @@
...
 		</div>
+		<div>
+			<h3>
+				<strong><?php p($l->t('Memory limits'));?></strong>
+			</h3>
+			<p class="settings-hint"><?php p($l->t('We limit the use of RAM according to your hardware, but you can change it. Higher memory may improve the results a bit, but it will also be unnecessarily slower.'));?>
+				<a target="_blank" rel="noreferrer noopener" class="icon-info" title="<?php p($l->t('Open Documentation'));?>" href="https://github.com/matiasdelellis/facerecognition/wiki/Performance-analysis-of-DLib%E2%80%99s-CNN-face-detection"></a>
+			</p>
+			<p>
+				<span><input type="memory" id="memory-limits-text" name="memory-limits-text" placeholder="<?php p($l->t('Use subfix as 2048 MB or 2G'));?>"></span>
```
stalker314314: maybe mention here that 4GB is maximum?

templates/settings/admin.php
```diff
@@ -25,6 +25,20 @@
 				<a id="save-sensitivity" class="icon-align icon-edit" style="display: none;" title="<?php p($l->t('Save'));?>" href="#"></a>
 			</p>
 		</div>
+		<div>
+			<h3>
+				<strong><?php p($l->t('Memory limits'));?></strong>
+			</h3>
+			<p class="settings-hint"><?php p($l->t('We limit the use of RAM according to your hardware, but you can change it. Higher memory may improve the results a bit, but it will also be unnecessarily slower.'));?>
```
stalker314314: ...but you can *limit it even more below configured PHP limits*.

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
matiasdelellis: I thought it but preferred to make the smallest patch, but now I do it. :wink:

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
matiasdelellis: Oh..  You're right, did not think well it, just copy other debug. :sweat_smile:

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
matiasdelellis: > if you don't use this value if $status is STATE_ERROR, why do you set it here then?

I could take it into account in the fronted as the suggested value.. But It just does not bother, the important thing is to indicate that the configuration is invalid.

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
matiasdelellis: I removed this. See if it's okay.

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
matiasdelellis: Done..

templates/settings/admin.php
matiasdelellis: Well, Also mention the minimum, that by default half of the RAM is used. There are many things to explain. Let's leave it on the wiki..  :wink:

matiasdelellis:
Hi, @stalker314314 
Please, A second review. Do not be afraid to insist on things that you already commented- :wink:

lib/Controller/SettingController.php
```diff
@@ -118,23 +120,45 @@ public function getUserValue($type) {
...
+				$this->config->setAppValue('facerecognition', $type, $value);
 				$this->userManager->callForSeenUsers(function(IUser $user) {
 					$this->config->setUserValue($user->getUID(), 'facerecognition', 'recreate-clusters', 'true');
 				});
 				break;
+			case 'memory-limits':
+				if (is_numeric ($value)) {
+					// Apply prundent limits.
+					if ($value < 1 * 1024 * 1024 * 1024) {
+						$value = 1 * 1024 * 1024 * 1024;
```
matiasdelellis: You asked why send this? The idea is that it can be used as valid value suggestions.

I'll leave it like that, unless they're incongruous, but this led me to think that we can add a parameter 'message', where explicitly explain the error, and show it in the notification. What you think?

Here:
> $messge = "The recommended analysis needs at least 1 Gb of RAM";

stalker314314: message is fine to me

matiasdelellis: Done.. Last review and merge it..

PR #161, opened 2019-09-09
Uses its own dialog to rename people.

matiasdelellis:
Well,
After so much time it is the only interesting thing I could do. :sweat_smile: 

 * Use a own dialog, based on OC.Dialog
 * Show image to whom you are editing.
 * Fill in the entry, with the original name.

How it looks .. 
Own-dialog

P.S.: It seems that we are in the stage in which "meets our expectations", and maybe a bit stagnants. Should not we publish for feedback?

### DIFF
--- a/appinfo/app.php
+++ b/appinfo/app.php
@@ -29,6 +29,7 @@ $app = new Application();
 
 \OC::$server->getEventDispatcher()->addListener('OCA\Files::loadAdditionalScripts', function(){
 	OC_Util::addVendorScript('facerecognition', 'js/lozad');
+	\OCP\Util::addScript('facerecognition', 'fr-dialogs');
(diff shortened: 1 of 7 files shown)


stalker314314:
> Should not we publish for feedback?

I have no problem with that! My wishlist (that I never get to do it) is to wrap up documentation (wiki) pages, maybe point to them directly from app (I like how PhoneTrack/GpxPod guy did it, not necessarily UI, but I think it is nice to go from app directly to wiki/github/issues...). All other stuff (except [these 2](https://github.com/matiasdelellis/facerecognition/issues?q=is%3Aopen+is%3Aissue+milestone%3A%22NC+app+store+publish%22)) are in "enhancement" category. I don't think we will be any closer to "perfect" than now, so go for it:)

Issue #143, opened 2019-05-09
Clusters on personal view should be sorted by their sizes


stalker314314:
@matiasdelellis - Yes, I know this is temporary view and I am aware this is not something we should invest a lot into:) That being said...I am very annoyed that people I care most (people having most faces in my collection) are not at the top. So, I propose this issue - sort persons by sizes of their respective clusters.
Now, I am completely fine that I do this, just wants an "ack" from you that you are OK with this work at all.

I was thinking to either:
1) add a new field `$cluster['faces_count']` in `PersonController::index` and let JS do the sorting, or
2) actually pre-sort clusters (server-side) during PersonMapper call

I would do 2) and maybe _even_ try to merge whole of `PersonController::index` into only one SQL call (so we don't do N lookup for faces in cluster). Thoughts?

matiasdelellis:
@stalker314314 

> Yes, I know this is temporary view and I am aware this is not something we should invest a lot into:)

Something that works a lot, but unfortunately with few results, is to make a virtual view in the file manager -simil Trash, o Share with... view-, that would be a good replacement for this view.. So after we publish I'll ask the Nextcloud guys for help..

Regarding this report:

> ...sort persons by sizes of their respective clusters.
>
> Now, I am completely fine that I do this, just wants an "ack" from you that you are OK with this work at all.

I agree with that.

> I was thinking to either:
>    add a new field $cluster['faces_count'] in PersonController::index and let JS do the sorting, or
>    actually pre-sort clusters (server-side) during PersonMapper call
>
>I would do 2) and maybe even try to merge whole of PersonController::index into only one SQL call (so we don't do N lookup for faces in cluster). Thoughts?

I guess the best would be a combination of both, to be able to add a simple button to select the ordering.

Issue #113, opened 2019-01-29
Do not show clusters with 1 face in it


stalker314314:
Not sure if this is a bug or a feature:), but I don't see anything good that can came out of this:) 40% of my clusters are clusters with 1 face and that can save some scrolling on page, thumbnail generation and user intimidation with lot of results thrown at him/her.

![slika](https://user-images.githubusercontent.com/188751/51941132-bd427480-240b-11e9-8ad4-78ff832ce7e5.png)

james-cook:
Just getting to know the system but I quite like to have even these clusters just to get a feeling of what's going on - especially when adjusting sensitivity. 
Maybe a filter to reveal/hide single member clusters and default set at hidden...

PR #162, opened 2019-09-10
Add option to not show single clusters... and sort by the size of the entire group ..

matiasdelellis:
* Clusters on personal view should be sorted by their sizes #143
* Do not show clusters with 1 face in it #113

### DIFF
--- a/lib/Controller/PersonController.php
+++ b/lib/Controller/PersonController.php
@@ -62,6 +62,7 @@ class PersonController extends Controller {
 	 */
 	public function index() {
 		$model = intval($this->config->getAppValue('facerecognition', 'model', AddDefaultFaceModel::DEFAULT_FACE_MODEL_ID));
+		$notGrouped = $this->config->getAppValue('facerecognition', 'show-not-grouped', 'false');
 		$userEnabled = $this->config->getUserValue($this->userId, 'facerecognition', 'enabled', 'false');
 
 		$resp = array();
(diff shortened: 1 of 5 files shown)


lib/Controller/PersonController.php
```diff
@@ -71,20 +72,30 @@ public function index() {
...
-				$cluster = [];
+				$personFaces = $this->faceMapper->findFacesFromPerson($this->userId, $person->getId(), $model);
+				if ($notGrouped === 'false' && count($personFaces) <= 1)
+					continue;
+
+				$limit = 14;
 				$faces = [];
-				$personFaces = $this->faceMapper->findFacesFromPerson($this->userId, $person->getId(), $model, 14);
 				foreach ($personFaces as $personFace) {
+					if ($limit-- === 0)
```
matiasdelellis: This is quite horrible, but the query is practically just as fast, and by ignoring the single groups, the query gets smaller.

templates/settings/admin.php
```diff
@@ -39,14 +39,31 @@
...
-		<h3><strong><?php p($l->t('Current status'));?></strong></h3>
 		<div>
-			<p id="progress-text"><?php p($l->t('Stopped'));?></p>
-			<progress id="progress-bar" value="0" max="100"></progress>
+			<h3>
+				<strong><?php p($l->t('Misc settings'));?></strong>
+			</h3>
+			<p>
+				<input id="showNotGrouped" name="showNotGrouped" type="checkbox" class="checkbox">
+				<label for="showNotGrouped"><?php p($l->t('Show people who are not grouped'));?></label><br>
```
stalker314314: #consider "Show person with only 1 face found"

matiasdelellis: Show persons with a only 1 face found ??

PR #165, opened 2019-10-08
Allow Nc17 tested on home server. Update travis matrix to test from N…

matiasdelellis:
Allow Nc17 tested on home server. Update travis matrix to test from NC14 to NC17

Wow...
We should test if it is still functional in NC14.. :sweat_smile:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,11 +18,17 @@ env:
     - APP_NAME=facerecognition
   matrix:
     - DB=mysql
-    - DB=mysql CORE_BRANCH=stable15
     - DB=pgsql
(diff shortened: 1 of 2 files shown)


matiasdelellis:
Hi @stalker314314 

https://github.com/nextcloud/server/wiki/Maintenance-and-Release-Schedule

We cannot ensure that it works in NC 14, and they no longer have maintenance. Do you think it's okay to support NC> = 16?

stalker314314:
sorry for missing this! (you should just go ahead and pull, if you see me gone like this next time:D)

PR #177, opened 2019-11-18
Set person as valid after cluster recreation, fixes #163

### LINKED ISSUE #163: person.'is_valid' is never reseted until recreate cluster
Hi @stalker314314 

How to reproduce:

 * For any reason a image is invalidated, the file changed or is removed.
 * It set the person.is_valid = FALSE for the **WHOLE PERSONS** in these image.

I still can see this Person in the group view of user settings page. MM.. This is a bug? For a single photo, what change should I stop showing everything?. In this case, i guess this is the right behavior .. I must show it.

However, when you click on any face of it person, it shows the full image, and when you see the persons in it, it does not show the invalid person.

I guess I should show all people too there.. No?

So, When is a person really invalid and should be ignored?

Returning to the report, the bug really is that when the task is executed, if the resulting cluster is the same -In the sense of the faces what it contains are the same-, the cluster is not rebuilt -Which is logical- but it is not validated again.

Then, I execute the task continuously and always reports:

```6/10 - Executing task CreateClustersTask (Create new persons or update existing persons)
	Found 0 faces without associated persons for user matias and model 1
	Found 1 changed persons for user matias and model 1
	1995 faces found for clustering
	986 persons found after clustering
```

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/80194371-person-is_valid-is-never-reseted-until-recreate-cluster?utm_campaign=plugin&utm_content=tracker%2F74944432&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F74944432&utm_medium=issues&utm_source=github).
</bountysource-plugin>

stalker314314:
When we were doing cluster recreation, only new clusters were getting
`is_valid` set as true. Existing (but changed) clusters was not getting
updated value for `is_valid`, so if it was false, it remained false.
This triggered new cluster recreations (as logic was thinking there are
stale persons and cluster recreation is needed) in infinite loop.

This change adds update to DB, so that each modified cluster is set as
valid.

Tests are also modified so they actually test this behavior.

### DIFF
--- a/lib/Db/PersonMapper.php
+++ b/lib/Db/PersonMapper.php
@@ -51,7 +51,7 @@ class PersonMapper extends QBMapper {
 
 	public function findAll(string $userId): array {
 		$qb = $this->db->getQueryBuilder();
-		$qb->select('id', 'name')
+		$qb->select('id', 'name', 'is_valid')
 			->from('face_recognition_persons', 'p')
 			->where($qb->expr()->eq('user', $qb->createNamedParameter($userId)));
 
@@ -216,6 +216,15 @@ class PersonMapper extends QBMapper {
 				foreach ($newFaces as $newFace) {
 					$this->updateFace($newFace, $newPerson);
 				}
+
+				// Set cluster as valid now
+				$qb = $this->db->getQueryBuilder();
+				$qb
+					->update($this->getTableName())
+					->set("is_valid", $qb->createParameter('is_valid'))
+					->where($qb->expr()->eq('id', $qb->createNamedParameter($newPerson)))
+					->setParameter('is_valid', true, IQueryBuilder::PARAM_BOOL)
+					->execute();
 			}
 
 			// Add new clusters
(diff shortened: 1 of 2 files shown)


matiasdelellis:
It seems to work .. :smiley: 

But I had people without faces again.. (Issue https://github.com/matiasdelellis/facerecognition/issues/137) But I do not associate it with this PR .. :sweat_smile:

matiasdelellis:
@stalker314314 
It is definitely related with #137 but as far as this PR is concerned, it meets the objective, therefore merge it and we discuss #137 on #137 again. :sweat_smile:

stalker314314:
hmm...why is there no Travis here, in checks? This build passed, but I expected it to guard bad PRs from getting pulled? Or... it was never here?:)

PR #178, opened 2019-11-19
Deletes persons without faces on change, fixes #137

### LINKED ISSUE #137: I have orphan persons (without faces), and always recreates the group.

stalker314314:
Using approach #1 from #137.

This change removes all persons that are without faces, after image
(file) is modified/deleted in hooks. When file is modified/deleted, we
are invalidating all persons for which there are faces on that image and
then removing faces. If this is last face for a given person, we didn't
delete that person, but we should (as there is no point having it), so
this fixed that.

### DIFF
--- a/lib/Watcher.php
+++ b/lib/Watcher.php
@@ -172,7 +172,17 @@ class Watcher {
 			// note that invalidatePersons depends on existence of faces for a given image,
 			// and we must invalidate before we delete faces!
 			$this->personMapper->invalidatePersons($imageId);
+
+			// Fetch all faces to be deleted before deleting them, and then delete them
+			$facesToRemove = $this->faceMapper->findByImage($imageId);
 			$this->faceMapper->removeFaces($imageId);
+
+			// If any person is now without faces, remove those (empty) persons
+			foreach ($facesToRemove as $faceToRemove) {
+				if ($faceToRemove->getPerson() !== null) {
+					$this->personMapper->removeIfEmpty($faceToRemove->getPerson());
+				}
+			}
 		}
 	}
 
@@ -226,10 +236,20 @@ class Watcher {
 			// note that invalidatePersons depends on existence of faces for a given image,
 			// and we must invalidate before we delete faces!
 			$this->personMapper->invalidatePersons($imageId);
+
+			// Fetch all faces to be deleted before deleting them, and then delete them
+			$facesToRemove = $this->faceMapper->findByImage($imageId);
 			$this->faceMapper->removeFaces($imageId);
 
 			$image->setId($imageId);
 			$this->imageMapper->delete($image);
+
+			// If any person is now without faces, remove those (empty) persons
+			foreach ($facesToRemove as $faceToRemove) {
+				if ($faceToRemove->getPerson() !== null) {
+					$this->personMapper->removeIfEmpty($faceToRemove->getPerson());
+				}
+			}
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


lib/Watcher.php
```diff
@@ -172,7 +172,17 @@ public function postWrite(Node $node) {
...
 			// and we must invalidate before we delete faces!
 			$this->personMapper->invalidatePersons($imageId);
+
+			// Fetch all faces to be deleted before deleting them, and then delete them
+			$facesToRemove = $this->faceMapper->findByImage($imageId);
 			$this->faceMapper->removeFaces($imageId);
+
+			// If any person is now without faces, remove those (empty) persons
+			foreach ($facesToRemove as $faceToRemove) {
+				if ($faceToRemove->getPerson() !== null) {
```
matiasdelellis: In that circumstance this would Null?. While in the background task the face is discovered and grouped?

matiasdelellis:
Hi @stalker314314 

Except that doubt this is fine ..

I was thinking of doing a function in imageMapper or some new service:
```
public function invalidateImage(int $imageId, bool detele) {
    // Put all the functions of the hooks that are similar here.
    if (delete === true) {
        // This is the only difference between both hooks..
        $this->imageMapper->delete($image);
    }
}
```

More than anything for cleaning .. but the current is completely acceptable.. :wink:

p.s: I was also seeing how to use transactions, but I hardly found its use in official Nextcloud projects,

matiasdelellis:
p.s: If you want to start approach 2, but as I said in #137  without that the database being invalid, it would be great. Meanwhile I think we can accept this.

About our doubts of the concurrence perhaps we should wait for experts from NC.

lib/Watcher.php
stalker314314: well, what if you add image, do face detection and there is no clustering yet (still not enough images...). One way to reach this situation!:)

stalker314314:
Approach 2 should be trivial. I will work on it later, but it is just as a "safety net". Should not block and release to app store though!

> p.s: I was also seeing how to use transactions

Actually, we are using transaction on couple of places already. In `ImageMapper::imageProcessed` and in `PersonMapper::mergeClusterToDatabase`! These parts needs to be done atomically

> but as I said in #137 without that the database being invalid, it would be great

Our states are by design invalid from time to time and we need to live with it (because detection takes so long and is not proactively done). Every time you add or modify image, you are in invalid state:) Only if we can somehow run face detection on any addition/modification of image, then we might have valid state always. But, today - we have to live with it:)

PR #179, opened 2019-11-19
Deleting orphaned persons from create cluster task

stalker314314:
Just to be sure there is no leftovers (persons without any faces), we do
another cleanup of orphaned persons, for good measure:)

(approach 1 remains)

### DIFF
--- a/lib/BackgroundJob/Tasks/CreateClustersTask.php
+++ b/lib/BackgroundJob/Tasks/CreateClustersTask.php
@@ -202,7 +202,10 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 
 		// Remove all orphaned persons (those without any faces)
 		// NOTE: we will do this for all models, not just for current one, but this is not problem.
-		$this->personMapper->deleteOrphaned($userId);
+		$orphansDeleted = $this->personMapper->deleteOrphaned($userId);
+		if ($orphansDeleted > 0) {
+			$this->logInfo('Deleted ' . $orphansDeleted . ' persons without faces');
+		}
 
 		// Prevents not create/recreate the clusters unnecessarily.
 		$this->config->setUserValue($userId, 'facerecognition', 'recreate-clusters', 'false');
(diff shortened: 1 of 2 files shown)


lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -200,6 +200,10 @@ private function createClusterIfNeeded(string $userId) {
 		$mergedClusters = $this->mergeClusters($currentClusters, $newClusters);
 		$this->personMapper->mergeClusterToDatabase($userId, $currentClusters, $mergedClusters);
 
+		// Remove all orphaned persons (those without any faces)
+		// NOTE: we will do this for all models, not just for current one, but this is not problem.
+		$this->personMapper->deleteOrphaned($userId);
```
matiasdelellis: Why not print a warning when delete orphaned persons?
Hopefully it will never be printed. but if it happens, we can analyze it ..
Just:
`if ($this->personMapper->deleteOrphaned($userId))`
or 
`if (count($this->personMapper->deleteOrphaned($userId))`

lib/Db/PersonMapper.php
```diff
@@ -290,6 +290,25 @@ public function removeIfEmpty(int $personId) {
 			->execute();
 	}
 
+	/**
+	 * Deletes all persons that have no faces associated to them
+	 *
+	 * @param string $userId ID of user for which we are deleting orphaned persons
+	 */
+	public function deleteOrphaned(string $userId) {
```
matiasdelellis: `public function deleteOrphaned(string $userId): int {`

lib/Db/PersonMapper.php
```diff
@@ -290,6 +290,25 @@ public function removeIfEmpty(int $personId) {
...
+		$sub->select(new Literal('1'));
+		$sub->from("face_recognition_faces", "f")
+			->where($sub->expr()->eq('f.person', 'p.id'));
+
+		$qb = $this->db->getQueryBuilder();
+		$qb->delete($this->getTableName(), 'p')
+			->where($qb->expr()->eq('p.user', $qb->createParameter('user')))
+			->andWhere('NOT EXISTS (' . $sub->getSQL() . ')')
+			->setParameter('user', $userId)
+			->execute();
```
matiasdelellis: The delete return an count of afected rows.

```
		$qb->delete($this->getTableName(), 'p')
			->where($qb->expr()->eq('p.user', $qb->createParameter('user')))
			->andWhere('NOT EXISTS (' . $sub->getSQL() . ')')
			->setParameter('user', $userId);
		return db->execute();
```

stalker314314:
Take a look now:)

Issue #181, opened 2019-11-21
Travis fail on master since sinse PR #179


matiasdelellis:
Why didn't we see it in the PR? :confused: 

The function that fail.

https://github.com/matiasdelellis/facerecognition/blob/86e5f044da7b8212ac2c14706f300cae2c27a236/lib/Db/PersonMapper.php#L298-L310

The query that use doctrine is:
```
DELETE FROM `oc_face_recognition_persons` p WHERE (`p`.`user` = ?) AND (NOT EXISTS (SELECT 1 FROM `oc_face_recognition_faces` `f` WHERE `f`.`person` = `p`.`id`)
```

... and with a quick investigation, at least in mariadb must be:
```
DELETE p FROM `oc_face_recognition_persons` p
WHERE (`p`.`user` = 'matias') AND
(NOT EXISTS (SELECT 1 FROM `oc_face_recognition_faces` `f` WHERE `f`.`person` = `p`.`id`))
```

PR #182, opened 2019-11-21
Fix deleteOrphaned on mysql. Issue #181

matiasdelellis:
It's ugly, but I know it works in mariadb/mysql :disappointed:

### DIFF
--- a/lib/Db/PersonMapper.php
+++ b/lib/Db/PersonMapper.php
@@ -297,7 +297,7 @@ class PersonMapper extends QBMapper {
 	 */
 	public function deleteOrphaned(string $userId): int {
 		$sub = $this->db->getQueryBuilder();
-		$sub->select('*');
+		$sub->select(new Literal('1'));
 		$sub->from("face_recognition_faces", "f")
 			->where($sub->expr()->eq('f.person', 'p.id'));
 
--- a/lib/Db/PersonMapper.php
+++ b/lib/Db/PersonMapper.php
@@ -297,16 +297,26 @@ class PersonMapper extends QBMapper {
 	 */
 	public function deleteOrphaned(string $userId): int {
 		$sub = $this->db->getQueryBuilder();
-		$sub->select(new Literal('1'));
+		$sub->select('*');
 		$sub->from("face_recognition_faces", "f")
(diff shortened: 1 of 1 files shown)


matiasdelellis:
Well, I received an email from Travis that it had failed when it had not yet started. :sweat_smile: 
But finally it pass correctly.

https://travis-ci.org/matiasdelellis/facerecognition/builds/615052529?utm_medium=notification&utm_source=email

lib/Db/PersonMapper.php
```diff
@@ -297,16 +297,26 @@ public function removeIfEmpty(int $personId) {
 	 */
 	public function deleteOrphaned(string $userId): int {
 		$sub = $this->db->getQueryBuilder();
-		$sub->select(new Literal('1'));
+		$sub->select('*');
```
stalker314314: any reason why you needed "*" here? you should keep "1" unless there is reason not to.

stalker314314:
I commented couple of days ago "https://github.com/matiasdelellis/facerecognition/pull/177#issuecomment-555235688_", like we lost Travis. This is probably why I didn't see build failure, Sorry for creating problem.

BTW, how is this working for FaceMapper::deleteUserFaces which uses similar construct:
https://github.com/matiasdelellis/facerecognition/blob/master/lib/Db/FaceMapper.php#L165

I vaguely remember we already had similar issues (but with SQLite...). Anyway, I approved this, shoot when ready!:)

lib/Db/PersonMapper.php
matiasdelellis: No reason ... it was just one of many tests. :sweat_smile:

matiasdelellis:
> BTW, how is this working for FaceMapper::deleteUserFaces which uses similar construct:
https://github.com/matiasdelellis/facerecognition/blob/master/lib/Db/FaceMapper.php#L165

I don't know.. . As I am understanding now mysql has conflicts with some nested queries, and doctrine with NOT EXISTS in some versions..
That's why I take what was working, and did the delete separately. It was the only thing that worked for me.. But I can't defend this change ..

matiasdelellis:
Well, with last commit pass again.. So, merge it.. Then we can improve it ..

PR #186, opened 2019-11-21
Trying out better description

stalker314314:


stalker314314:
@matiasdelellis I think we should mention that 4GB of RAM is required. Maybe even slip that additional installation is required, like "Once you install all required software...". If we don't mention that, I think it would be dishonest from out side and it can cause frustration/bad experience for people who install it and cannot use it. What do you think? What else I missed?

matiasdelellis:
About the current description is fine.. :smile: 

> @matiasdelellis I think we should mention that 4GB of RAM is required. Maybe even slip that additional installation is required, like "Once you install all required software...". If we don't mention that, I think it would be dishonest from out side and it can cause frustration/bad experience for people who install it and cannot use it. What do you think? What else I missed?

You're right.. I don't know if say a number (I leave it to your critic), but we must say that It is a service with high requirements, which works progressively, probably slow, and it is essential to read the documentation. It should be clear that just installing it does nothing. :sweat_smile: 

So, don't have to alarm so much, but you have to advice to users. :+1:

stalker314314:
> In case it is not clear, approved ..

It was clear, but I am getting to this repo only in the evenings:)

Please check again! Added warnings

appinfo/info.xml
```diff
@@ -5,6 +5,10 @@
 	<summary>A face recognition app</summary>
 	<description><![CDATA[**Detect and group faces of your loved one in your cloud**
 
+⚠️ This application requires minimum of 12GB of RAM memory to work! See [Requirements](https://github.com/matiasdelellis/facerecognition/wiki/Requirements-and-Limitations) for details.
```
matiasdelellis: 12GB????
I have only 8.. :disappointed:

appinfo/info.xml
```diff
@@ -5,6 +5,10 @@
...
+
+⚠️ Setup of this app requires access to terminal and even getting your hands dirty with installation of additional software. See [Installation](https://github.com/matiasdelellis/facerecognition/wiki/Installation) for details.
```
matiasdelellis: :+1:

appinfo/info.xml
stalker314314: OMG, good you noticed:) I didn't copy "-" nicely:)

Issue #150, opened 2019-06-12
Bug report template..


matiasdelellis:
Hi @stalker314314 
This issue is to discuss a bug reporting template to be used by users.
Basically it is to suggest a basic information that they must provide so that we can help them more efficiently.

See:
https://github.blog/2016-02-17-issue-and-pull-request-templates/

That's the Nextcloud server template.. Modify it here to fit our needs.

Template:
========

Hey, Thanks for reporting issues back to Nextcloud Face Recognition. Try filling this report with the greatest care so that we can help you. :smile: 

Make sure you read all the [documentation](https://github.com/matiasdelellis/facerecognition/wiki), and the [FAQ](https://github.com/matiasdelellis/facerecognition/wiki/FAQ), and that the issue has not been reported before. :wink: 

### Expected behaviour
Tell us what should happen

### Actual behaviour
Tell us what happens instead

### Steps to reproduce
1.
2.
3.

### Server configuration

**Operating system**:

**Pdlib version:**

**How build dlib:**

**How build pdlib:**

**PHP version:**

**Web server:**

**Database:**

**Nextcloud version:**

### Client configuration
**Browser:**

**Operating system:**

#### Background task log
<details>
<summary>sudo -u apache php occ face:background_job</summary>

```
Insert your background log here
```
</details>

#### Web server error log
<details>
<summary>Web server error log</summary>

```
Insert your webserver log here
```
</details>

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>

```
Insert your Nextcloud log here
```
</details>

#### Browser log
<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```
</details>

stalker314314:
Looks fine to me:)

PR #187, opened 2019-11-22
Issue template

matiasdelellis:
Add the issue template.. Issue #150 

I never add it because I intended to require some basic test to the report.

Now publish this test here:
 * https://github.com/matiasdelellis/pdlib-min-test-suite

It needs to be improve, but  wth that can easily test dlib and pdlib instalation

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,4 +1,4 @@
-Hey, Thanks for reporting issues back to Nextcloud Face Recognition. Try filling this report with the greatest care so that we can help you. :smile:
+Hey, Thanks for reporting issues back to Nextcloud Face Recognition. Please, try to complete this report in detail so we can help you easier. :smile:
 
 Make sure you read all the [documentation](https://github.com/matiasdelellis/facerecognition/wiki), and the [FAQ](https://github.com/matiasdelellis/facerecognition/wiki/FAQ), and that the issue has not been reported before. :wink:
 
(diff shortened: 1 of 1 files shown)


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,82 @@
+Hey, Thanks for reporting issues back to Nextcloud Face Recognition. Try filling this report with the greatest care so that we can help you. :smile:
```
stalker314314: I would remove "greatest". It seems like we are some corporation:) It can kind-of be of putting. If I see this, I would say "screw you, I want to help you and you are intimidating me":)... but, maybe it is just me:)

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,82 @@
...
+2.
+3.
+
+### Server configuration
+
+* **Operating system:**
+
+* **Pdlib version:**
+
+* **How installed dlib:** Make sure it is installed correctly with https://github.com/matiasdelellis/pdlib-min-test-suite
```
stalker314314: "How is DLib installed" seems more natural?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,82 @@
...
+
+* **How installed pdlib:** Make sure it is installed correctly with https://github.com/matiasdelellis/pdlib-min-test-suite
```
stalker314314: Can you place link as a [link](link)

stalker314314:
You wanted separate repo on purpose? You didn't want to taint this repo, or you though it would be easier for people to check-out separate repo? Latter makes sense to me, but just wanted to confirm

ISSUE_TEMPLATE.md
matiasdelellis: haha. Ok.. :sweat_smile:

matiasdelellis:
> You wanted separate repo on purpose? You didn't want to taint this repo, or you though it would be easier for people to check-out separate repo? Latter makes sense to me, but just wanted to confirm

The applications installed from the app store generally do not include any tests.
The idea is to have something very small that any person can download to ensure that it is working. I did it separately, because it is easier to download the repository from there, cloning, or just downloading a compressed file.

On the other hand, yes, we can dirty that repository without any regrets. :sweat_smile: 

But the main thing for me is to help/support of dlib/pdlib problems there... and here try to talk only about the application. :wink:

ISSUE_TEMPLATE.md
matiasdelellis: Please, try to complete this report in detail so we can help you easier. :smile:
¿?

Issue #154, opened 2019-07-03
Error: Call to a member function getParent() on bool (When viewing clusters)


qosmio:
## Configuration information

**Pdlib Version: 0.1.0
Models Version: 1
Nextcloud: 16.0.1
OS: Ubuntu 18.04
DB: Postgres 11.4
Browser: Firefox Quantum 64
PHP: 7.3.4
Web Server: NGINX 1.16
Facerecognition Commit: cc43fa5c62f08ae170d13fba867c7b76cccdaddc
Dlib: 19.17**

I'm having an issue viewing the clusters page, even though I've run a full scan for my own username. I've actually never been able to view the clusters page, even on version 15.

![Annotation](https://user-images.githubusercontent.com/8892380/60623405-2e65c980-9da8-11e9-9f5c-14c3d5e235ec.png)

Below is the error 
### PHP Error
```php

[index] Error: Error: Call to a member function getParent() on bool at <<closure>>

0. /data/code/nextcloud/apps/facerecognition/lib/Controller/PersonController.php line 80
   OCA\FaceRecognition\Controller\PersonController->getRedirectToFileUrl(45166)
1. /data/code/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 166
   OCA\FaceRecognition\Controller\PersonController->index()
2. /data/code/nextcloud/lib/private/AppFramework/Http/Dispatcher.php line 99
   OC\AppFramework\Http\Dispatcher->executeController(OCA\FaceRecognit ... {}, "index")
3. /data/code/nextcloud/lib/private/AppFramework/App.php line 126
   OC\AppFramework\Http\Dispatcher->dispatch(OCA\FaceRecognit ... {}, "index")
4. /data/code/nextcloud/lib/private/AppFramework/Routing/RouteActionHandler.php line 47
   OC\AppFramework\App::main("OCA\\FaceRecogn ... r", "index", OC\AppFramework\ ... {}, {_route: "facerecognition.person.index"})
5. <<closure>>
   OC\AppFramework\Routing\RouteActionHandler->__invoke({_route: "facerecognition.person.index"})
6. /data/code/nextcloud/lib/private/Route/Router.php line 297
   undefinedundefinedcall_user_func(OC\AppFramework\ ... {}, {_route: "facerecognition.person.index"})
7. /data/code/nextcloud/lib/base.php line 975
   OC\Route\Router->match("/apps/facerecognition/persons")
8. /data/code/nextcloud/index.php line 42
[...]

PR #192, opened 2019-11-28
Small fixes and improvements ..

matiasdelellis:
- Add button to show all clusters with same person name. [1]
- Add a button to go back and show all clusters. Before you had to update the entire page. downloading the clusters again.
- Force delete invalid entries when a file does not exist. Old issue #154
- Try to be consistent in that we work with clusters at least in api/endpoint's. 

**Screenshots**
imagen
imagen

p.s: It is a small way to suggest to users that there will be many groups of the same person, without going into technical matters.

### DIFF
--- a/lib/Controller/PersonController.php
+++ b/lib/Controller/PersonController.php
@@ -23,6 +23,7 @@ use OCA\FaceRecognition\Db\Person;
 use OCA\FaceRecognition\Db\PersonMapper;
 
 use OCA\FaceRecognition\Migration\AddDefaultFaceModel;
+use OCA\FaceRecognition\BackgroundJob\Tasks\StaleImagesRemovalTask;
 
 class PersonController extends Controller {
 
@@ -84,8 +85,10 @@ class PersonController extends Controller {
 
 					$image = $this->imageMapper->find($this->userId, $personFace->getImage());
 					$fileUrl = $this->getRedirectToFileUrl($image->getFile());
-					if (NULL === $fileUrl)
+					if (NULL === $fileUrl) {
+						$limit++;
 						continue;
+					}
 
 					$face = [];
 					$face['thumb-url'] = $this->getThumbUrl($personFace->getId());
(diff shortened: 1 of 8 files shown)


matiasdelellis:
p.s: Also select the name when open the rename dialog.. I AM VERY HAPPY WITH THIS TONTERY! :sweat_smile:

lib/Controller/PersonController.php
```diff
@@ -69,46 +70,49 @@ public function index() {
...
 		$resp['clusters'] = array();
 
-		if ($userEnabled === 'true') {
-			$persons = $this->personMapper->findAll($this->userId);
-			foreach ($persons as $person) {
-				$personFaces = $this->faceMapper->findFacesFromPerson($this->userId, $person->getId(), $model);
-				if ($notGrouped === 'false' && count($personFaces) <= 1)
-					continue;
+		if ($userEnabled !== 'true')
+			return new DataResponse($resp);
```
stalker314314: :thumbsup:

lib/Controller/PersonController.php
```diff
@@ -69,46 +70,49 @@ public function index() {
...
 
-				$cluster = [];
-				$cluster['name'] = $person->getName();
-				$cluster['count'] = count($personFaces);
-				$cluster['id'] = $person->getId();
-				$cluster['faces'] = $faces;
+				$image = $this->imageMapper->find($this->userId, $personFace->getImage());
+				$fileUrl = $this->getRedirectToFileUrl($image->getFile());
+				if (NULL === $fileUrl) {
+					$limit++;
```
stalker314314: consider moving `if ($limit-- === 0)` at the end, instead of doing ++/-- with $limit:)

lib/Controller/PersonController.php
```diff
@@ -161,8 +209,10 @@ private function getRedirectToFileUrl($fileId) {
 		$files      = $baseFolder->getById($fileId);
 		$file       = current($files);
 
-		if(!($file instanceof File))
+		if(!($file instanceof File)) {
+			$this->config->setUserValue($this->userId, 'facerecognition', StaleImagesRemovalTask::STALE_IMAGES_REMOVAL_NEEDED_KEY, 'true');
```
stalker314314: maybe write some comment why is this key needed (and what `$file` can be if not `File`). I think it might be beneficial to "future us":)

lib/Db/PersonMapper.php
```diff
@@ -49,6 +48,15 @@ public function find(string $userId, int $personId): Person {
 		return $person;
 	}
 
+	public function findByName(string $userId, string $personName): array {
```
stalker314314: consider exercising this method in `MergeClusterToDatabaseTest` instead of some of those `findAll` methods (should be relatively easy to just change `findAll` with N `findByName`

matiasdelellis: I'm still trying to fix travis .. I don't want to change more things .. :disappointed:

Issue #193, opened 2019-11-29
Initial face preview generation hang Nextcloud


Petahh:
Hello All, when I access the recognized face those little square previews of faces start to load for the very first time, often the whole instance of Nextcloud freeze.

I'm using debian/apache version of Nextcloud with MariaDB, Redis with 4GB of RAM.

So I found that when I go line by line it loads in mannered fashion, but If I maximize and it tries to load all of the previews at once (for the first time) it works only unlit eats all the memory then hangs. Is there any way how to pace these previews or pre-generate all manually?

matiasdelellis:
Hello @Petahh 
You're right, today we don't have any cache implementation.

We originally intended to use the browser cache, but having the faces of all your photos in the browser can be questionable..

As you observed, they are dynamically loaded as you move the viewport, but the faces are so small that you easily have 100 faces in sight at all times (This is 100 calls in parallel to your server). and if you scroll to the end of the page, it will try to load all your faces, which can be an important load for your server.

In principle, I never wanted to cache the server side because I think that this view should not exist (It would be for a long discussion :wink: ), and writing millions of small images on your disk, also would be an important load for your server, since writing to disk is even more expensive..

Well, we currently use [Lozad.js](https://github.com/ApoorvSaxena/lozad.js) to load the images dynamically. My first idea would be to look for an alternative that limits the loading of images to a certain number in parallel, but of course it is not a real solution to the problem.

> it works only unlit eats all the memory then hangs.

Sorry for that! :sweat:

Petahh:
No worries @matiasdelellis , I managed to generate all by slowly scrolling down in small window. Now it works like charm. It’s something which definitely needs to be improved for future.

matiasdelellis:
Although it is difficult, if it advances, we can do something similar ..

https://github.com/nextcloud/server/pull/18210

GAS85:
Why not to use standard Previews folder of Nextcloud and just added there face1.jpg, face2.jpg? There is a 1 Folder per picture with different previews.

matiasdelellis:
Hi @Petahh @GAS85 
I have been working to eliminate the face view, in favor of an photos view reusing Nextcloud thumbnails, and it could dramatically help with this problem..  See # 433, I would appreciate tests and comments.. :smile:

GAS85:
Could't find anything under #433...

matiasdelellis:
Sorry.. 😅 
[...]

PR #194, opened 2019-11-30
Implement an basic cache for faces.. Issue #193

matiasdelellis:
Please @Petahh

Small attempt to improve your report.. They are not yet pre-generated, but you can verify that once generated, the load of your server should be lighter to show a complete view.

### DIFF
--- a/lib/Controller/FaceController.php
+++ b/lib/Controller/FaceController.php
@@ -17,11 +17,14 @@ use OCA\FaceRecognition\Db\FaceMapper;
 use OCA\FaceRecognition\Db\Image;
 use OCA\FaceRecognition\Db\ImageMapper;
 
+use OCA\FaceRecognition\Service\FileCache;
+
 class FaceController extends Controller {
 
 	private $rootFolder;
 	private $faceMapper;
 	private $imageMapper;
+	private $fileCache;
 	private $userId;
 
 	public function __construct($AppName,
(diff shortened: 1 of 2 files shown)


matiasdelellis:
@stalker314314 
I add that they are pregenerated after saving the faces to the database? 
However, it worries me that it would be a much larger patch than it seemed..
I mean, I have no problem pregenerating it -Compared to the discovery process, it must be much faster..-, but I don't want to have to eliminate it every time something happens with an image/face, ..

matiasdelellis:
*Note:* Nextcloud never eliminates the preview cache ... and that cache would be much bigger ... Can we live with this?

Petahh:
I think only problem is initial pre-generation. If you have many pictures. Maybe could be part of the background job we run to analyze and cluster faces?

stalker314314:
I *can* live if you want to implement this. I think it will be super-hard to nail this right, with all needed cache evictions and so on. But if you want to try it - go for it!

I am more wondering is this problem *we* need to solve? Thumbs are 32x32, right? I think that [Thumbnail Generator](https://apps.nextcloud.com/apps/previewgenerator) app is already doing some thumbnail generation, and even default settings would suite us, right? (32x32 should be one of default sizes for `squareSizes`) Then, that app takes care of all gory details of what it is doing best - thumbnail generation, and we take care of what we are best:) We should add wiki page explaining and pointing people to Thumbnail generator in FAQ, but that's it - preview generation should not be our job and we should not be wasting time on problem of thumbnail generator, cache population, cache evictions... That's my 2c. If [Thumbnail Generator](https://apps.nextcloud.com/apps/previewgenerator) is not helping, we can discuss is this patch "worth it", to say it like that:)

matiasdelellis:
Hi both,

> I think only problem is initial pre-generation. If you have many pictures. Maybe could be part of the background job we run to analyze and cluster faces?

I agree, maybe do the pre-generation, but until we have an easy cleaning this will not be merged..

> I can live if you want to implement this. I think it will be super-hard to nail this right, with all needed cache evictions and so on. But if you want to try it - go for it!

As I said, generate them is easy, but I don't want to deal with cleaning, it must be something more or less automatic.
The implementation of the garbage collector is simple, but it does not work due it depends on the modification date, and not on the last access date. IAppData storage is [SimpleFS](https://github.com/nextcloud/server/tree/master/lib/public/Files/SimpleFS), that does not support touch to update the modification date and therefore it is a dead end .. :disappointed: 

> I am more wondering is this problem we need to solve?

If it were just a performance problem, I wouldn't worry about fixing it. but the problem in #193 is that the number of calls in parallel to get the faces breaks the server..

Perhaps it is just an Apache problem, which should not accept so many queries in parallel, and should be limit the threads. I do not know. :confused: 

> Thumbs are 32x32, right? I think that Thumbnail Generator app is already doing some thumbnail generation, and even default settings would suite us, right? (32x32 should be one of default sizes for squareSizes) Then, that app takes care of all gory details of what it is doing best - thumbnail generation, and we take care of what we are best:) We should add wiki page explaining and pointing people to Thumbnail generator in FAQ, but that's it - preview generation should not be our job and we should not be wasting time on problem of thumbnail generator, cache population, cache evictions... That's my 2c. If Thumbnail Generator is not helping, we can discuss is this patch "worth it", to say it like that:)

The previewgenerator app create the complete thumbnail of files, and you can indicate the size, but not the position (Faces position..), or name of file... It does not help us ..

Biont:
Would it not help immensely already if you set appropriate headers for client side caching?
```php
$dataDisplayResponse->cacheFor($days * 24 * 60 * 60);
$dataDisplayResponse->addHeader('Pragma', 'cache');
return $dataDisplayResponse;
```

matiasdelellis:
Hi @Biont 
Thanks for the suggestion, it's something that I need to fix, but I wasn't convinced to cache all the faces in the browser..  but if it improve the behavior I can accept a patch with this change .. :wink:

PR #195, opened 2019-12-03
Fix Travis

matiasdelellis:
…aceRecognitionContext::__construct()

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,10 +51,10 @@ before_install:
 
 before_script:
   - ./occ app:check-code $APP_NAME || true
-  - ./occ face:setup
   - cd apps/$APP_NAME
 
 script:
+  - ../../occ face:setup
   - make test
 
   # Create coverage report
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,13 +48,18 @@ before_install:
   - bash ./before_install.sh $APP_NAME $CORE_BRANCH $DB
   - cd ../server;
   - ./occ app:enable $APP_NAME
+  - ./occ face:setup
 
 before_script:
   - ./occ app:check-code $APP_NAME || true
   - cd apps/$APP_NAME
 
 script:
-  - bash ./tests/travis-script.sh
+  - make test
+
+  # Create coverage report
+  - wget https://scrutinizer-ci.com/ocular.phar
+  - php ocular.phar code-coverage:upload --format=php-clover clover.xml
 
 after_failure:
   - cat ../../data/nextcloud.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -56,9 +56,5 @@ before_script:
 script:
   - bash ./tests/travis-script.sh
 
-  # Create coverage report
-  - wget https://scrutinizer-ci.com/ocular.phar
-  - php ocular.phar code-coverage:upload --format=php-clover clover.xml
-
 after_failure:
   - cat ../../data/nextcloud.log
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,7 @@ before_script:
   - cd apps/$APP_NAME
 
 script:
-  - ../../occ face:setup && make test
+  - bash ./tests/travis-script.sh
 
   # Create coverage report
   - wget https://scrutinizer-ci.com/ocular.phar
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,10 +48,10 @@ before_install:
   - bash ./before_install.sh $APP_NAME $CORE_BRANCH $DB
   - cd ../server;
   - ./occ app:enable $APP_NAME
-  - ./occ face:setup
 
 before_script:
   - ./occ app:check-code $APP_NAME || true
+  - ./occ face:setup
   - cd apps/$APP_NAME
 
 script:
@@ -63,4 +63,3 @@ script:
 
 after_failure:
   - cat ../../data/nextcloud.log
-
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,13 +48,13 @@ before_install:
   - bash ./before_install.sh $APP_NAME $CORE_BRANCH $DB
   - cd ../server;
   - ./occ app:enable $APP_NAME
+  - ./occ face:setup
 
 before_script:
   - ./occ app:check-code $APP_NAME || true
   - cd apps/$APP_NAME
 
 script:
-  - ./occ face:setup
   - make test
 
   # Create coverage report
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,8 +54,7 @@ before_script:
   - cd apps/$APP_NAME
 
 script:
-  - ../../occ face:setup
-  - make test
+  - ../../occ face:setup && make test
 
   # Create coverage report
   - wget https://scrutinizer-ci.com/ocular.phar
--- a/lib/BackgroundJob/Tasks/ImageProcessingTask.php
+++ b/lib/BackgroundJob/Tasks/ImageProcessingTask.php
@@ -151,6 +151,8 @@ class ImageProcessingTask extends FaceRecognitionBackgroundTask {
 		$dataDir = rtrim($context->config->getSystemValue('datadirectory', \OC::$SERVERROOT.'/data'), '/');
 		$images = $context->propertyBag['images'];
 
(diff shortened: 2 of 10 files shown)


matiasdelellis:
@stalker314314 
Not approve of this, because it does not work and I was out of ideas .. :sweat_smile: 

It seems that ModelService cannot find the models ..

stalker314314:
> Not approve of this, because it does not work and I was out of ideas .. 

No, I am approving this *in general*, so you don't wait for me;) Whole PR is fine.

tests/unit/RequirementsTest.php
```diff
@@ -56,6 +56,13 @@ public function setUp() {
 		$this->context->logger = new FaceRecognitionLogger($logger);
 	}
 
+	public function testModelInstalled() {
+		$config = $this->createMock(IConfig::class);
+		$defaultModel = AddDefaultFaceModel::DEFAULT_FACE_MODEL_ID;
+		$modelInstalled = intval($config->getAppValue('facerecognition', 'model', AddDefaultFaceModel::DEFAULT_FACE_MODEL_ID));
```
matiasdelellis: @stalker314314 
Last travis says:
 * `Failed asserting that 0 matches expected 1.`

So, https://github.com/matiasdelellis/facerecognition/blob/master/lib/Migration/AddDefaultFaceModel.php is not executed??

stalker314314: if it was not installed, all tests would fail? And `getAppValue` would return provided default value, which is `AddDefaultFaceModel::DEFAULT_FACE_MODEL_ID`, so it should match even if repair step is not run:)

There is this line in Travis:
```
Build done. You can enable the application in Nextcloud.
./vendor/bin/phpunit --coverage-clover clover.xml -c phpunit.xml
```

Do Travis need to enable app? Only after app is enabled, upgrade step should be run?

Weird...

stalker314314: I never used `$config` as mock in unit tests, only in integration tests. Are you sure mock object is not just returning default value for int type, which is 0?

matiasdelellis:
Ok,,
Experimenting a bit, I see that the model is installed in:
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620418632#L1168
 * `Decompresing: /tmp/oc_tmp_YpNvPv-folder-.facerecognition/dlib_face_recognition_resnet_model_v1.dat.bz2 on /home/travis/build/matiasdelellis/server/data-autotest/appdata_oc4mp74c3vge/facerecognition/models/1/dlib_face_recognition_resnet_model_v1.dat`

And open the same model file.
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620418632#L1340
 * `Exception: Unable to open /home/travis/build/matiasdelellis/server/data-autotest/appdata_oc4mp74c3vge/facerecognition/models/1/mmod_human_face_detector.dat for reading.`

But the file does not exist ...

bypassed here: https://github.com/matiasdelellis/facerecognition/blob/b84ebe7b27ab88884862fad1fea042426c22b70e/lib/Helper/Requirements.php#L61

:confused:

.travis.yml
```diff
@@ -54,13 +54,7 @@ before_script:
   - cd apps/$APP_NAME
 
 script:
-  - ./occ face:setup
-  - make test
-
-  # Create coverage report
-  - wget https://scrutinizer-ci.com/ocular.phar
-  - php ocular.phar code-coverage:upload --format=php-clover clover.xml
+  - bash ./tests/travis-script.sh
```
matiasdelellis: I move everything within the same script to ensure run within the same 'context' or 'stage'

lib/Command/SetupCommand.php
```diff
@@ -91,7 +91,7 @@ protected function execute(InputInterface $input, OutputInterface $output) {
 		$requirements = new Requirements($this->modelService, $this->modelVersion);
 		if ($requirements->modelFilesPresent()) {
 			$this->logger->writeln('The models are already installed');
-			return 0;
```
matiasdelellis: HACK: modelFilesPresent() now is hacked to always print fullpath, so it is always true.. So, force reinstall here..

lib/Command/SetupCommand.php
```diff
@@ -201,6 +201,8 @@ private function bunzip2 ($in, $out) {
 
 		bzclose ($in_file);
 		fclose ($out_file);
+
+		$this->logger->writeln(filesize($out)/(1024*1024) . ' Mb');
```
matiasdelellis: Just print size to ensure that the file was decompresed.

lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -151,6 +151,8 @@ public function execute(FaceRecognitionContext $context) {
 		$dataDir = rtrim($context->config->getSystemValue('datadirectory', \OC::$SERVERROOT.'/data'), '/');
 		$images = $context->propertyBag['images'];
 
+		$this->logInfo('File ' . $requirements->getFaceDetectionModel() . ': '. filesize($requirements->getFaceDetectionModel())/(1024*1024) . ' Mb');
```
matiasdelellis: Try to print size to compare it and make sure that file exists ..
Here's where it fails.. or if I delete this line, it fails when try to open the model.

lib/Helper/Requirements.php
```diff
@@ -58,7 +58,7 @@ public function getFaceDetectionModel() {
 	private function getModel1File(string $file) {
 		$fullPath = $this->modelService->getModelPath($file);
 
-		if (file_exists($fullPath)) {
+		if (true || file_exists($fullPath)) {
```
matiasdelellis: Bypass to always return fullpath.. Otherwise, when you try to open the model, the file will be null.

tests/travis-script.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+../../occ face:setup
```
matiasdelellis: Install the models.. and it is verified that it exists with the size ..

tests/travis-script.sh
```diff
@@ -0,0 +1,8 @@
...
+make test
```
matiasdelellis: It does the tests, and they fail because the files don't exist anymore. :confused: :confounded: :sob: 

@stalker314314 Do you have any idea?

stalker314314: no. Which files? I think you are too deep in this at this point:)

matiasdelellis: :sweat_smile: 
Add comments to the commit to explain a bit, but don't focus on it.

See travis.. it is easier to see the problem ..

Well, the model is installed..
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620595745#L1253

..but when trying to use them he doesn't find them ..
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620595745#L1342

stalker314314: this is weirdest thing I am seeing recently:) I have no idea...can you try doing `ls -alR` sprinkled all in `make test` Makefile, so we see if when it is missing? At beginning, at end and in between all of these `test-bin-deps vendor-deps js-templates` targets

matiasdelellis: Model folder is empty...
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620659199#L1288

Install models... 
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620659199#L1293

Model folder now has the models..
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620659199#L1350

Build the app (Since here don't install the models anymore maybe is not needed...): This should not affect appdata..
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620659199#L1357

Run tests ..
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620659199#L1420

It fails because it does not exist .. :disappointed: 
 * https://travis-ci.org/matiasdelellis/facerecognition/jobs/620659199#L1430

Appdata is empty.. :confused: 
https://travis-ci.org/matiasdelellis/facerecognition/jobs/620659199#L1467

matiasdelellis: https://github.com/nextcloud/travis_ci/blob/master/core_install.sh#L108

:disappointed:

stalker314314: omfg, lol
So...tests need to install models?

matiasdelellis: > ~https://github.com/nextcloud/travis_ci/blob/master/core_install.sh#L108~

I'm still wrong.. This command is executed before..

Issue #26, opened 2018-10-04
Support for shared storage


stalker314314:
`Watcher.php` has some TODO for this (`ISharedStorage`). But yeah, in general, we want to know are we shared-storage friendly, or not.

stalker314314:
Just to add here, since I bumed into this recently - seems like these guys know how to deal with both external/group folders, we can _steal_ a thing of two from them: https://github.com/nextcloud/files_fulltextsearch/tree/master/lib/Service

urlyb:
Thanks for the great app. So, if understand correctly, NC "External storages" are not supported yet?

stalker314314:
Yes, you are correct. This app doesn't (yet) support neither shared nor external storage. I will surely work on external first, as this is where my photos are:) However, current thinking is that we want to release on store what we have up to now, and then continue from there, so expect something probably summer 2019 or so. Unless, you want to deal with it yourself, in which case, you are more than welcome!

migueltg:
Hi, first of all, thanks for the app. One question, current first release supports external storage?

stalker314314:
> 
> 
> Hi, first of all, thanks for the app. One question, current first release supports external storage?

No, unfortunately no. We plan to work on that next (at least I plan to, as this is my pain point, most of my photos are on external storage...)

migueltg:
> > Hi, first of all, thanks for the app. One question, current first release supports external storage?
> 
> No, unfortunately no. We plan to work on that next (at least I plan to, as this is my pain point, most of my photos are on external storage...)

Thanks for your quick reply and your effort. If i can help with anything, tell to me :)

matiasdelellis:
@stalker314314 
I've been thinking about this for quite a while ... and it became clear when I investigated to replace private api. Issue #191

Our limitation of shared folders is due pdlib needs physical access to the files..?
However, we don't work with real files either. We always create temporary files (With size according to memory allocation), and with PDlib we read these files..
So why not use the IPreview interface to get these temproral files?.

stalker314314:
Your approach looks like it *could* work:) Here are my biggest pain points (let's use this issue for discussion, feel free to move this to #198 if you think it is more appropriate)
1. Does your PR handle external storage?
2. How to scope it only on files in shared folders (and external storage), not on trash, or some third-party app images (we should be careful what we are detecting:)
[...]

PR #198, opened 2019-12-06
Shared storage experiments.. :bomb: 

matiasdelellis:
@stalker314314 
This is just an experiment. 
You are the most interested so I share it, and I explain a little... but I will not continue. :wink: :see_no_evil: 

I have a folder with The big bang Theory images in another instance..

imagen

As you can see in the sidebar i have it shared with my instance..

imagen

And tada, my instance is analyzing the files. :open_mouth:

The idea is taken from the preview interface, and is enough generic to work with any storage.. :smile: 

Well, Maybe I'm being very innocent about it .. :sweat_smile: 

I explain a little more inside the commit.. :wink:

### DIFF
--- a/lib/BackgroundJob/Tasks/AddMissingImagesTask.php
+++ b/lib/BackgroundJob/Tasks/AddMissingImagesTask.php
@@ -33,7 +33,6 @@ use OCA\FaceRecognition\BackgroundJob\FaceRecognitionBackgroundTask;
 use OCA\FaceRecognition\BackgroundJob\FaceRecognitionContext;
 use OCA\FaceRecognition\Db\Image;
 use OCA\FaceRecognition\Db\ImageMapper;
-use OCA\FaceRecognition\Helper\Requirements;
 use OCA\FaceRecognition\Migration\AddDefaultFaceModel;
 use OCA\FaceRecognition\Service\FileService;
 
@@ -142,7 +141,7 @@ class AddMissingImagesTask extends FaceRecognitionBackgroundTask {
 	 */
 	private function parseUserFolder(string $userId, int $model, Folder $folder): int {
 		$insertedImages = 0;
-		$nodes = $this->getPicturesFromFolder($folder);
+		$nodes = $this->fileService->getPicturesFromFolder($folder);
 		foreach ($nodes as $file) {
 			$this->logDebug('Found ' . $file->getPath());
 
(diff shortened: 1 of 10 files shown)


lib/BackgroundJob/Tasks/AddMissingImagesTask.php
```diff
@@ -164,13 +163,7 @@ private function parseUserFolder(int $model, Folder $folder): int {
 	 * @return array List of all images and folders to continue recursive crawling
 	 */
 	private function getPicturesFromFolder(Folder $folder, $results = array()) {
-		// todo: should we also care about this too: instanceOfStorage(ISharedStorage::class);
```
matiasdelellis: Easy.. No?
I just ignore any control and accept all the files. 

Well, If we want to make it more serious, we have a problem, if you print the class of the shared folder, does this say TRASH ??? (We had already seen this :confused:)

But in all the implementations that I saw, (RichDocuments, Gallery, Core) do not control the type of storage .. Just if is local..

lib/BackgroundJob/Tasks/AddMissingImagesTask.php
```diff
@@ -134,14 +133,14 @@ private function addMissingImagesForUser(string $userId, int $model): int {
...
 	 */
-	private function parseUserFolder(int $model, Folder $folder): int {
+	private function parseUserFolder(string $userId, int $model, Folder $folder): int {
 		$insertedImages = 0;
 		$nodes = $this->getPicturesFromFolder($folder);
 		foreach ($nodes as $file) {
 			$this->logDebug('Found ' . $file->getPath());
 
 			$image = new Image();
-			$image->setUser($file->getOwner()->getUid());
```
matiasdelellis: If I take the user here, this is matiasdelellis@cloud-******. is that of the other instance, so use the userId from addMissingImagesForUser().

If it is shared in our instance, we must consider it local, and each user individually.

My only doubts here, is if we can analyze the photos that someone share directly with us. Perhaps it would not be necessary to analyze when the original user does not activate it (or even installed as in this case.), but it would be a very fine discussion. In practice, if someone shares a file, you can use it however you want... :see_no_evil: 

How does the saying ...? if you can see it you can copy it.. :sweat_smile:

lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -383,4 +384,30 @@ private function calculateMaxImageArea(): int {
...
 		return $maxImageArea;
 	}
+
+	/**
+	 * Get a path to either the local file or temporary file
+	 *
+	 * @param File $file
+	 * @param int $maxSize maximum size for temporary files
+	 * @return string
+	 */
```
matiasdelellis: This function is taked from IPreview\ProviderV2 that is new but all internal functions exists since OC 9

https://github.com/nextcloud/server/blob/12bdb95c1a40170652570decef7f466e4b97aa65/lib/private/Preview/ProviderV2.php#L81

As it comment, it returns the local file path, or a temporary file path downloaded from the original instance. :grimacing: 

..and this is all the magic .. You can read the files of any storage. :smile:

lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -212,7 +212,8 @@ private function findFaces(\CnnFaceDetection $cfd, string $dataDir, Image $image
 		}
 
 		// todo: this concat is wrong with shared files.
-		$imagePath = $dataDir . $file[0]->getPath();
```
matiasdelellis: Note that everything can be cleaner.

matiasdelellis:
p.s: Here I said that we can use the IPreview interface to obtain temporary files, but as it is here it is unnecessary. (and it is not desirable, since we would create large files that would never be deleted.)

https://github.com/matiasdelellis/facerecognition/issues/26#issuecomment-562604765

matiasdelellis:
p.s: It caught my attention that the clusters/faces view worked correctly.. ..and is that we are using the same function getLocalFile() :sweat_smile: 

https://github.com/matiasdelellis/facerecognition/blob/3f9681c639a1e7feb7696f01da1965060f943369/lib/Controller/FaceController.php#L58

matiasdelellis:
@stalker314314 
As comment earlier, this branch ended up being the introduction of a FileService, to simplify a lot of code.
The support of shared files is quite proven, but will remain experimental, hidden to admin/users.. Note that also automatically gain support of encrypted files. (See issue #201 ) :grimacing: 

After merging this, I will add the side panel to disable folders. and surely then it introduces a SettingsService where to simplify access to configurations and variables. :wink:

Issue #171, opened 2019-11-11
Feature: custom exclusion folder


millionart:
I have some special folders for storing landscapes and emoticons. However, these images are still mixed with facial recognition results. I understand that facial recognition results are temporarily imperfect, so I hope to allow me to manually disable certain images from being analyzed.

This exclusion function is best to use the folder (total exclusion) and the label (individual exclusion) function.

stalker314314:
Hi,
FaceRecognition do have something like what you need - if you place empty file in folder, named `.nomedia`, it will exclude that folder, as well as all children folders from face detection, completely. Furthermore, I think that application is smart enough that, moment that you place this file in some foldee, all images/faces under that folders will be deleted and clustering rescheduled! Would that work for you (to exclude folders with landscapes/emoticons)?

What is "label" (individual exclusion)? Excluding individual files? FaceRecognition definitely don't have that, not sure what @matiasdelellis thinks of adding it. To me, doesn't look like big issue and I am not even sure how to implement it without some custom logic.

P.S. I just realized that nowhere in wiki we are talking about this feature, so I'll add it - keep this issue active.

millionart:
Thank you! It works!

stalker314314:
I added explanation to:
https://github.com/matiasdelellis/facerecognition/wiki/Usage
https://github.com/matiasdelellis/facerecognition/wiki/FAQ
Will close task now:)

berho:
Hello,
first of all: many thanks for the great job done on facerecognition!! I am on the way to run over > 300 GB of images (7 days left ;)

But I am not happy with the ".nomedia" file: I have folders containing lot of images I do not want to recognice faces on it (public events) but I want to view them with Gallery app! But when I place a file named ".nomedia" inside the folder it is not visible to Gallery anymore!

matiasdelellis:
Ok @berho 
It seems a real case to improve .. :smile: 

I can't find the discussion, but at some point we talked about using a `.nofacemedia` file with the same purpose. 
It doesn't convince me 100%, but I suppose it is an easy solution to a real problem.

stalker314314:
Maybe we can make it app (or user) config? Defaults to `.nomedia`, but user can change it to whatever they want (maybe keep it hidden and require to be a file)?

matiasdelellis:
I am not sure about the name of the file but I would not make it configurable. 
[...]

PR #202, opened 2019-12-18
Add the side panel to the folders to enable or disable them

matiasdelellis:
Fix issue #171 

## Enabled..
imagen

## Disabled
imagen

## Not supported..
imagen

### DIFF
--- a/js/files-tabview.js
+++ b/js/files-tabview.js
@@ -6,7 +6,8 @@
 
         events: {
             'click .icon-more': '_onPersonClickMore',
-            'click a.icon-rename': '_onRenamePerson'
+            'click a.icon-rename': '_onRenamePerson',
+            'click #searchPersonsToggle': '_onSearchPersonsToggle'
         },
 
         getLabel: function() {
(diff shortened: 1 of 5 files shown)


js/files-tabview.js
```diff
@@ -21,99 +22,164 @@
...
+        getFolderTemplate: function (data) {
+            var html = "";
+            html += "<div class='emptycontent'>";
+            html += "<div class='icon-user svg'></div>";
+            html += "<p>";
+            html += "<input class='checkbox' id='searchPersonsToggle'";
+            if (data.child_detection)
+                html += "checked='checked'";
+            html += "type='checkbox'>";
+            html += "<label for='searchPersonsToggle'>" + t('facerecognition', 'Search for my loved ones in this folder.') + "</label>";
```
stalker314314: this "loved ones" seems very patronizing to me (and can be far out of context). I would keep it neutral ("search for faces"/"apply face recognition"/...)

js/files-tabview.js
```diff
@@ -21,99 +22,164 @@
...
+            html += "</p>";
+            html += "<span>" + t('facerecognition', 'Folders with <em>.nomedia</em> files still will be ignored.') + "</span>";
```
stalker314314: Folders with <em>.nomedia</em> file will remain unaffected by this.

lib/Service/FileService.php
```diff
@@ -138,6 +146,32 @@ public function allowsChildDetection(Folder $folder): bool {
 		return true;
 	}
 
+	/**
+	 * Set this folder to enable or disable the analysis using the .facerecognition.json file.
+	 *
+	 * @param Folder $folder Folder to enable/disable for
+	 * @return bool true if the change is done. False if failed.
+	 */
+	public function setAllowChildDetection(Folder $folder, bool $detection): bool {
```
stalker314314: Ha ha, maybe change to DescendantDetection or something, this is creepy:)

js/files-tabview.js
stalker314314: maybe warn user that new file will be created in this directory?

js/files-tabview.js
matiasdelellis: haha.. :sweat_smile: 
I always thought the same .. but I had based on google photos.

matiasdelellis: Maybe .. 'Search for persons in the images of this directory.'

js/files-tabview.js
matiasdelellis: MM.. 
I would rather not be so technical.  Do not even notify about .nomedia file.

Maybe: "If the images are outside the gallery they will be ignored."

PR #203, opened 2019-12-19
Yet another attempt to fix travis

matiasdelellis:


matiasdelellis:
Wooooowww...
A little ugly, but this works .. :sweat_smile:

PR #222, opened 2020-01-28
Try to be more consistent with memory requirements.

matiasdelellis:
The limit of 1GB of minimum ram is real.. . and not allow its execution.
Reduce the minimum memory to 300MB, that theoretically would be an image
area of 640x480px.
Use centralized settings in SettingsService, and although it is not
recommended they are easily editable.

### DIFF
--- a/lib/BackgroundJob/Tasks/ImageProcessingTask.php
+++ b/lib/BackgroundJob/Tasks/ImageProcessingTask.php
@@ -388,7 +388,7 @@ class ImageProcessingTask extends FaceRecognitionBackgroundTask {
 		// Based on amount on memory PHP have, we will determine maximum amount of image size that we need to scale to.
 		// This reasoning and calculations are all based on analysis given here:
 		// https://github.com/matiasdelellis/facerecognition/wiki/Performance-analysis-of-DLib%E2%80%99s-CNN-face-detection
-		$maxImageArea = intval((0.75 * $allowedMemory) / 1024); // in pixels^2
+		$maxImageArea = intval(($allowedMemory) / 1024); // in pixels^2
 
 		return $maxImageArea;
 	}
--- a/lib/BackgroundJob/Tasks/ImageProcessingTask.php
+++ b/lib/BackgroundJob/Tasks/ImageProcessingTask.php
@@ -385,10 +385,11 @@ class ImageProcessingTask extends FaceRecognitionBackgroundTask {
 		// Calculate it from memory
 		//
 		$allowedMemory = $this->context->propertyBag['memory'];
+
 		// Based on amount on memory PHP have, we will determine maximum amount of image size that we need to scale to.
 		// This reasoning and calculations are all based on analysis given here:
 		// https://github.com/matiasdelellis/facerecognition/wiki/Performance-analysis-of-DLib%E2%80%99s-CNN-face-detection
-		$maxImageArea = intval(($allowedMemory) / 1024); // in pixels^2
+		$maxImageArea = intval(($allowedMemory) / SettingsService::MEMORY_AREA_RELATIONSHIP); // TODO: Maybe another helper.
 
 		return $maxImageArea;
 	}
--- a/lib/Service/SettingsService.php
+++ b/lib/Service/SettingsService.php
@@ -34,7 +34,7 @@ class SettingsService {
 	/*
 	 * System
 	 */
-	const MINIMUM_SYSTEM_MEMORY_REQUERIMENTS = 2 * 1024 * 1024 * 1024;
+	const MINIMUM_SYSTEM_MEMORY_REQUIREMENTS = 2 * 1024 * 1024 * 1024;
 	const MEMORY_AREA_RELATIONSHIP = 4 * 1024; // TODO: Move to models when implement others.
 
 	/*
--- a/lib/Service/SettingsService.php
+++ b/lib/Service/SettingsService.php
@@ -34,7 +34,8 @@ class SettingsService {
 	/*
 	 * System
 	 */
-	const MINIMUM_SYSTEM_MEMORY_REQUERIMENTS = 1 * 1024 * 1024 * 1024;
+	const MINIMUM_SYSTEM_MEMORY_REQUERIMENTS = 2 * 1024 * 1024 * 1024;
+	const MEMORY_AREA_RELATIONSHIP = 4 * 1024; // TODO: Move to models when implement others.
 
 	/*
 	 * Settings keys and default values.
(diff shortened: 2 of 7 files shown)


lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -388,7 +388,7 @@ private function calculateMaxImageArea(): int {
 		// Based on amount on memory PHP have, we will determine maximum amount of image size that we need to scale to.
 		// This reasoning and calculations are all based on analysis given here:
 		// https://github.com/matiasdelellis/facerecognition/wiki/Performance-analysis-of-DLib%E2%80%99s-CNN-face-detection
-		$maxImageArea = intval((0.75 * $allowedMemory) / 1024); // in pixels^2
```
matiasdelellis: I have pending return this somewhere more coherent.

According to your study, the photo analysis, linearly requires 1GB of ram for each MB de area in px. Which is very easy, but when cut it here is difficult to evaluate.. 

I think that cutting this value is important only when we use the values calculated from the php.ini But in configured values we must trust them.

lib/Service/SettingsService.php
```diff
@@ -53,7 +58,7 @@ class SettingsService {
 
 	/** Memory limit suggested for analysis */
 	const MEMORY_LIMITS_KEY = "memory-limits";
-	const MINIMUM_MEMORY_LIMITS = 1 * 1024 * 1024 * 1024;
+	const MINIMUM_MEMORY_LIMITS = 300 * 1024 * 1024; // 300 MB ensures a minimum area of 640x480px
```
matiasdelellis: 640px * 480px = 307200 px2 
307200 px2/1024 = 300Kpx = 0.3Mpx (I have a unit problem, but it is understood!) :sweat_smile: 

According to this need 300mb of ram. :grimacing: 
https://github.com/stalker314314/dlib-cnn-face-detection-analysis/blob/master/assets/memory-vs-type.png

640x480 is a good value this size? I guess 90% of personal photos would have important faces in the foreground, and therefore it should work well. Well, although this may be small, is the minimum, not the recommended .. :wink: 

The problem until now is that we said we need 1GB of ram, and we also set 1GB for analysis, and although within the analysis we cut it to 750Mb, it didn't make sense.

stalker314314: 640x480 should be good for large faces. Only thing to watch out is whether 300MB is enough? I know I tested it and it was sort-of linear, but not sure if linearity is at edges (with very low or high values of pixel areas). If you tried it with 300MB cap and it worked - go for it!:)

matiasdelellis: I'm trying to investigate this .. :wink:

matiasdelellis: imagen

matiasdelellis: imagen

Okey.. According to my analysis, the relationship between the area of the image (px*px), and the memory used (byte) absolutely linear, but multiplied by 4. 

Well.. This relationship maybe can change according to Dlib compilation options. :confused:

Width | 200px | 300px | 400px | 500px | 600px | 700px | 800px | 900px | 1.000px | 1.100px | 1.200px | 1.300px
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Total Time | 14s | 29s | 47s | 72s | 103s | 138s | 179s | 224s | 278s | 329s | 391s | 463s
  |   |   |   |   |   |   |   |   |   |   |   |  
Area | 40.000 | 90.000 | 160.000 | 250.000 | 360.000 | 490.000 | 640.000 | 810.000 | 1.000.000 | 1.210.000 | 1.440.000 | 1.690.000
Max. Memory | 169.092,00 | 346.835,97 | 674.472,04 | 1.051.688,04 | 1.489.419,98 | 1.993.847,81 | 2.574.543,87 | 3.266.652,16 | 3.996.928,00 | 4.782.096,38 | 5.688.212,48 | 6.655.320,06
Avg. Time | 0,93 | 1,93 | 3,13 | 4,80 | 6,87 | 9,20 | 11,93 | 14,93 | 18,53 | 21,93 | 26,07 | 30,87
  |   |   |   |   |   |   |   |   |   |   |   |  
Memory/Area | 4,23 | 3,85 | 4,22 | 4,21 | 4,14 | 4,07 | 4,02 | 4,03 | 4,00 | 3,95 | 3,95 | 3,94


So, to get these values I areate a new user test, and add 15 random photos to nextcloud. Then I use [nedtada](https://github.com/netdata/netdata/) to take the memory consumption series from the php process using the following script...

```
[matias@nube nextcloud]$ systemctl stop httpd php-fpm #To avoid analyzing any other php process
[matias@nube nextcloud]$ cat test-facerecognition.sh 
#!/usr/bin/env bash
NETDATAHOST='http://localhost:19999'
NCUSER='test'
#TIMEWINDOW='10'

MINSIDE='200'
MAXSIDE='1500'
DELTASIDE='100'

for ((i=${MINSIDE} ; i < ${MAXSIDE} ; i+=${DELTASIDE})); do
	let area=$((i ** 2))

	echo
	echo Reseting Nextcloud user: ${NCUSER}
	echo

	sudo -u apache php occ face:reset_all -u ${NCUSER}

	echo
	echo Testing an maximum image area of ${area}
	echo

	AFTER=$(date +%s)
	sudo -u apache php occ face:background_job -u ${NCUSER} --max_image_area ${area} -vvv #-t ${TIMEWINDOW}
	BEFORE=$(date +%s)

	curl -s "${NETDATAHOST}/api/v1/data?chart=apps.mem&before=${BEFORE}&after=${AFTER}&dimensions=php" | jq '.data[][1]'

	ALLTIME=$((${BEFORE}-${AFTER}))
	echo
	echo Take ${ALLTIME} seconds for ${area} of area
	echo

	sleep 3


```

matiasdelellis: Hi @stalker314314 

Well... Although weird, since I never complained, this is consistent with what I observed on my machine (Setting up 2GB the kernel always killed the process). :sweat_smile: 
...you ever verified this ..?

I will leave this value 4, as a new php defined parameter and I will calculate the area based on this relationship.
I insist that it can surely change according to the compilation, but certainly this seems to be the worst case. :disappointed: 

The problem now is that if we consider the minimum area of 640x480=0.3Mpx, according to this new calc, you need at least 1.2GB of RAM

Therefore I suggest that the minimum ram requirement is 2Gb

matiasdelellis: ...and if max memory is 4GB, so max area = 1Mpx :confused:

matiasdelellis:
So, with the last commit...
 * Reset user `user` where I have the pictures of Big Bang Theory.
 * Change the maximum memory parameter in settings.
 * Run the analysis on the photos..

..and repeating with several values, now see that the memory parameter is respected quite well.  :smile: 

Of course I insist that in other cases this may be different, and I would like other people to evaluate.

lib/Service/SettingsService.php
```diff
@@ -34,7 +34,8 @@ class SettingsService {
 	/*
 	 * System
 	 */
-	const MINIMUM_SYSTEM_MEMORY_REQUERIMENTS = 1 * 1024 * 1024 * 1024;
+	const MINIMUM_SYSTEM_MEMORY_REQUERIMENTS = 2 * 1024 * 1024 * 1024;
```
stalker314314: typo in requirmeents;D

PR #230, opened 2020-02-10
Add wipe clustering command on reset...

matiasdelellis:
Generalize the ResetAll command to add more options. Now is just Reset, and you must to pass --all argument to reset everything as until now, or --clustering to just drop clusters..

The command is now face:reset and need an argument about what you want to reset.
--all, to delete everything, or --clustering to reset persons.

### DIFF
--- a/lib/Command/ResetCommand.php
+++ b/lib/Command/ResetCommand.php
@@ -112,7 +112,8 @@ class ResetCommand extends Command {
 			return 0;
 		}
 		else {
-			$output->writeln('You must specify that you want to reset...');
+			$output->writeln('You must specify that you want to reset');
+			return 1;
 		}
 	}
 
--- a/lib/Command/ResetCommand.php
+++ b/lib/Command/ResetCommand.php
@@ -95,7 +95,8 @@ class ResetCommand extends Command {
 		if (!is_null($userId)) {
 			$user = $this->userManager->get($userId);
 			if ($user === null) {
-				throw new \InvalidArgumentException("User with id <$userId> in unknown.");
+				$output->writeln("User with id <$userId> in unknown.");
+				return 1;
 			}
 		}
 
--- a/lib/Command/ResetCommand.php
+++ b/lib/Command/ResetCommand.php
@@ -1,7 +1,7 @@
 <?php
 /**
- * @copyright Copyright (c) 2017, Matias De lellis <mati86dl@gmail.com>
- * @copyright Copyright (c) 2018, Branko Kokanovic <branko@kokanovic.org>
+ * @copyright Copyright (c) 2020, Matias De lellis <mati86dl@gmail.com>
+ * @copyright Copyright (c) 2019, Branko Kokanovic <branko@kokanovic.org>
  *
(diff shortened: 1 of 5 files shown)


lib/Db/FaceMapper.php
```diff
@@ -175,6 +198,26 @@ public function deleteUserFaces(string $userId) {
...
+	/**
+	 * Unset relation beetwen faces and persons from that user in order to reset clustering
+	 *
+	 * @param string $userId User to drop fo unset relation.
+	 */
+	public function unsetPersonsRelationForUser(string $userId) {
+		$sub = $this->db->getQueryBuilder();
+		$sub->select(new Literal('1'));
+		$sub->from('facerecog_images', 'i')
+			->where($sub->expr()->eq('i.id', '*PREFIX*' . $this->getTableName() .'.image'))
```
stalker314314: should you filter here by model? As a user, I would be annoyed if I do `reset` and it wipes clustering of my non-current model:)

lib/Db/FaceMapper.php
```diff
@@ -175,6 +198,26 @@ public function deleteUserFaces(string $userId) {
 			->execute();
 	}
 
+	/**
+	 * Unset relation beetwen faces and persons from that user in order to reset clustering
+	 *
+	 * @param string $userId User to drop fo unset relation.
+	 */
+	public function unsetPersonsRelationForUser(string $userId) {
```
stalker314314: I might be stupid, but I don't see point of setting this to NULL when you are calling `deleteUserPersons` after you call this? Resetting clustering is done by deleting all entries in `persons` table for a given user, why you need this?

lib/Command/ResetCommand.php
```diff
@@ -87,9 +101,27 @@ protected function execute(InputInterface $input, OutputInterface $output) {
...
+			$output->writeln('Reset successfully done');
+			return 0;
+		}
+		else if ($input->getOption('clustering')) {
+			$this->resetClusters($user);
+			$output->writeln('Reset clustering done');
+			return 0;
+		}
+		else {
+			$output->writeln('You must specify that you want to reset...');
```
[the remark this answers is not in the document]
matiasdelellis: MM.. No .. It is not relevant for any user... and probably either not for the administrator who installs the application. Maybe it is only useful for our test.. (Or maybe my experiments. haha).

lib/Db/FaceMapper.php
matiasdelellis: Ohh.. You're right!

lib/Db/FaceMapper.php
matiasdelellis: Honestly, just to make it more real. The command option resets the clusters... How is the database before clustering? It has the column `person` as NULL. Therefore, I tried to replicate this.

lib/Db/FaceMapper.php
matiasdelellis: The same applies to delete persons, but I am seeing that we (Probably just me) ignore the consideration of the model in many PersonMapper queries. :disappointed: 

That said, there will be a future PR when implement other models. :disappointed:

Issue #105, opened 2019-01-13
Edge calculation is very slow


stalker314314:
Edge calculation is done prior to giving edges to `dlib_chinese_whispers`, in this for loop:
https://github.com/matiasdelellis/facerecognition/blob/71c8238ba534e572fa9c26942078cab0ba9b2660/lib/BackgroundJob/Tasks/CreateClustersTask.php#L208-L219

However, I have, for example, 12000 faces and calculation takes 20-30 minutes, which is far more than 15 minutes that Nextcloud advises for cron jobs (and in some long future, we want to be able to finish any operation in 15 minutes). This is too long. Not sure what solution is, maybe we will need to copy array of all descriptors down to pdlib and do edge calculation there, we should try it out and see if that would speed it up.

I guess this is not for first version, maybe in future

matiasdelellis:
Hi @stalker314314 

I was playing with this (using foreach(), count() outside of for(), array_push(), array_merge(), references), but I not found great improvements.. :disappointed:

I only found one small detail...

https://github.com/matiasdelellis/facerecognition/blob/71c8238ba534e572fa9c26942078cab0ba9b2660/lib/BackgroundJob/Tasks/CreateClustersTask.php#L210

$j must be initialized in $i+1, to avoid adding ''$edges[] = array($i, $i);'

In your case, it saves 12000 comparisons... Maybe you win one second..  :sweat_smile:

stalker314314:
Nope, unfortunately, even that is not possible: https://github.com/davisking/dlib/blob/f7f6f6761817f2e6e5cf10ae4235fc5742779808/tools/python/src/face_recognition.cpp#L245 (maybe it is, but I don't want to "experiment":D)

matiasdelellis:
I seen it, but since use pre-increment (++i) there, I thought it worked like this... which is incorrect.. :disappointed: 
I think I remember that before, the behavior of the for() change if you used, pre or post increment.. :confused: 

Obviously, I'm wrong.. :sweat_smile: 
Sorry. Ignore my comments. :wink:

matiasdelellis:
Well... 73% of the time, it is due to the calculation of the Euclidean distance.

![valgrind](https://user-images.githubusercontent.com/733715/54860494-8f85f780-4cf9-11e9-8b47-e391907ac2d9.jpeg)

To analyze it yourself:
 * https://mike42.me/blog/2018-06-how-i-made-my-php-code-run-100-times-faster

[...]

PR #232, opened 2020-02-17
Implement a self contained class for Model code.

matiasdelellis:
I have this code hanging around for quite some time. :see_no_evil: 

Well .. This PR still has pending, but I think it's ready for a first review .. 

- [x] Implement and use an class which groups the dlib code..
- [x] Abtract classes to generalize CNN model.
- [x] Implement DlibCnn5Model (The current model) and DlibCnn68Model that extend it.
- [x] Implement an UGLY ModelManager which selects between both models.
- [x] face:setup can install both models.
- [ ] Abtract again in a generic Model class, that includes the ModelService code, and the models reuse them directly.
- [x] Use $model->isInstalled() and $model->meetDependencies()
- [ ] Drop Requeriments. Still manage MimeType handling.
- [x] The minimum configurations and the relation of the image vs memory depend on each model .. Therefore move it there.
- [x] Analyze memory consumption of 68 landmarks model..
- [x] HOG????
- [ ] Ensure that all functions properly handle the selected model.


lib/Service/SettingsService.php
```diff
@@ -43,18 +43,18 @@ class SettingsService {
...
 	const MINIMUM_SENSITIVITY = '0.4';
-	const DEFAULT_SENSITIVITY = '0.5';
+	const DEFAULT_SENSITIVITY = '0.4';
 	const MAXIMUM_SENSITIVITY = '0.6';
 
 	/** Minimum confidence used to try to clustring faces */
 	const MINIMUM_CONFIDENCE_KEY = 'min-confidence';
 	const MINIMUM_MINIMUM_CONFIDENCE = '0.0';
-	const DEFAULT_MINIMUM_CONFIDENCE = '0.9';
+	const DEFAULT_MINIMUM_CONFIDENCE = '0.99';
```
matiasdelellis: hahaha.. Well .. this escaped me, but it was for another PR. :sweat_smile:

matiasdelellis:
Ok.. HOG is working..  :grimacing: 

imagen

Unfortunately there are errors in the queries/controlles that must be fixed. For example, when the change the models, the persons query returns an empty list of persons without faces. This is due he simply ignores the model. See: https://github.com/matiasdelellis/facerecognition/blob/master/lib/Db/PersonMapper.php#L60
imagen

I'm afraid there are many queries where the same thing happens.. :disappointed: 
The problem that to make this query taking into account the model, you have to make an inner join between 3 tables. I think we get to the point that we need to refactor the tables, and add modelId to all tables exactly the same as userId. It would be disappointing, but the current structure of the table becomes unmanageable. :confused: 

@stalker314314 I need your opinion. Of course without hurry.. :wink: 

Regards,
Matias

matiasdelellis:
On the other hand .. First statistics .. :smile: 
And comparison with default model.. :wink: 

Area | 40000 | 90000 | 160000 | 250000 | 360000 | 490000 | 640000 | 810000 | 1000000 | 1210000 | 1440000 | 1690000
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Max. Memory (CNN) | 169.092,00 | 346.835,97 | 674.472,04 | 1.051.688,04 | 1.489.419,98 | 1.993.847,81 | 2.574.543,87 | 3.266.652,16 | 3.996.928,00 | 4.782.096,38 | 5.688.212,48 | 6.655.320,06
Avg. Time (CNN) | 0,93 | 1,93 | 3,13 | 4,80 | 6,87 | 9,20 | 11,93 | 14,93 | 18,53 | 21,93 | 26,07 | 30,87
Max Memory (HOG) | 81,48 | 77,53 | 81,81 | 78,46 | 82,58 | 70,94 | 79,98 | 103,64 | 84,90 | 85,89 | 78,79 | 87,76
AVG Time (HOG) | 1,31 | 2,25 | 3,13 | 3,56 | 5,56 | 7,19 | 7,25 | 11,31 | 13,56 | 16,38 | 16,13 | 22,50

imagen

We already know that memory consumption would be much better, but never imagine that it would simply be ridiculous.. :sweat_smile: 
It has no relationship between image size and memory consumption. Always kept below 100Mb 

About times theres are relatively short by little, but at first glance it seems more erratic.

In summary, memory is no longer a problem, but we cannot establish a direct relationship as with CNN. Well, although useless in terms of memory control, I think I continued using the same memory control, but with a ratio of 1Mpx = 1Gb ram for HOG. This will result in image analysis from 1.2Mpx to 8Mpx, and obviously this affects the results of the number of faces found ..

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
```diff
@@ -64,27 +73,28 @@ public function description() {
...
-				"Fill an issue here if that doesn't help: https://github.com/matiasdelellis/facerecognition/issues";
+		if (!$model->meetDependencies()) {
+			$error_message = "Seems that you don't meet the dependencies to use the model " . $modelId .": " . $model->getName();
+			// Document models on wiki and print link here.
 			$this->logInfo($error_message);
 			return false;
 		}
 
-		if (!$req->hasEnoughMemory()) {
+		$systemMemory = MemoryLimits::getSystemMemory();
```
stalker314314: hm, just a thought....maybe you can abstract away memory requirements under `meetDependecies`? Memory requirements are specific to model, right? HOG might not have memory limit to be checked, or some other model (imagine remote service as a model) will not have it. I would argue that memory requirements is just another requirements that is specific to model

lib/BackgroundJob/Tasks/ImageProcessingTask.php
```diff
@@ -389,7 +402,7 @@ private function calculateMaxImageArea(): int {
 		// Based on amount on memory PHP have, we will determine maximum amount of image size that we need to scale to.
 		// This reasoning and calculations are all based on analysis given here:
 		// https://github.com/matiasdelellis/facerecognition/wiki/Performance-analysis-of-DLib%E2%80%99s-CNN-face-detection
-		$maxImageArea = intval(($allowedMemory) / SettingsService::MEMORY_AREA_RELATIONSHIP); // TODO: Maybe another helper.
+		$maxImageArea = intval(($allowedMemory) / $model->getMemoryAreaRelation()); // TODO: Maybe another helper.
```
stalker314314: I would not expose `getMemoryAreaRelation` as something model have. I would rather expose something like `model::GetRecommendedImageArea` and keep memory are relation and such things deep in model logic. Some models will have memory are relations, some will not have it. Some will have bandwidth relation (remote service model) or something else, so maybe it makes sense to expose just what a model is "recommending" as max area

stalker314314:
First congratulation, this is huge (and beautiful change). I was thinking about this (for the sake of adding HOG), but your refactor is just...awesome, perfect!

> The problem that to make this query taking into account the model, you have to make an inner join between 3 tables. I think we get to the point that we need to refactor the tables, and add modelId to all tables exactly the same as userId. It would be disappointing, but the current structure of the table becomes unmanageable. 😕

Are you hitting some problem with schema where it is impossible to get model from some places? Or more like convenience problem where you don't want to join 3 tables? If it is convenience, I would be afraid to refactor just because of it (better would be to fix those wrong queries, right?)

> It has no relationship between image size and memory consumption. Always kept below 100Mb

I was always suspecting that HOG is O(1) for memory. Real question now is - what is your subjective feeling with HOG? If precision/recall is good enough (it recognizes lot of faces that CNN also do, and it doesn't recognize a lot of other objects that are not faces), maybe you should put it as default model? That will really, really make FaceRecognition app available for all! (only to figure out how not to compile dlib:D) We can get HOG from opencv, Chinese whispers might be rewritten in PHP, only missing part where we need dlib is to get face vectors:)

lib/BackgroundJob/Tasks/CheckRequirementsTask.php
matiasdelellis: > Memory requirements are specific to model, right?.

Is much more important the memory usage of model, but it still consumes a lot of memory when cluster the faces.. Do you remember your OOM when grouping with 0.6 sensitivity?

My idea was remove Requirements, but maybe I should keep it to leave this class cleaner.

lib/BackgroundJob/Tasks/ImageProcessingTask.php
matiasdelellis: I totally agree, but my doubts is the front. I can't generalize it so much. Maybe should remove it and ask questions in the setup command.
Ideally, send the path of the real image, and have the model resize it if necessary.. But I would leave it for another PR.

matiasdelellis:
> First congratulation, this is huge (and beautiful change). I was thinking about this (for the sake of adding HOG), but your refactor is just...awesome, perfect!

:sweat_smile: Thanks!.

> Are you hitting some problem with schema where it is impossible to get model from some places? Or more like convenience problem where you don't want to join 3 tables? If it is convenience, I would be afraid to refactor just because of it (better would be to fix those wrong queries, right?)

I agree that it would be better to use the current table, but there are an old saying.. "Normalize until it hurts, denormalize until it works." :sweat_smile: 
I wonder if we are at this point, because queries that are very simple, become very complicated. :confused: 

> I was always suspecting that HOG is O(1) for memory. Real question now is - what is your subjective feeling with HOG? If precision/recall is good enough (it recognizes lot of faces that CNN also do, and it doesn't recognize a lot of other objects that are not faces),

I still did not do real tests with my data, only with the Big Bang theory. Maybe I will do it later and make a comparison .. :wink: 

> maybe you should put it as default model? That will really, really make FaceRecognition app available for all! (only to figure out how not to compile dlib:D)

You're right... Maybe put it by default ..

> We can get HOG from opencv,

haha.. If anyone wants to contribute it would be great to also have openCV models, etc.  

> Chinese whispers might be rewritten in PHP, only missing part where we need dlib is to get face vectors:)

I doubt we can do that .. Remember this: https://github.com/matiasdelellis/facerecognition/issues/105

However, clustering could also be part of the model. I think it escapes this PR, but in the future it is possible ..
On the other hand, my next PR for pdlib will be just to compare the vectors there.

stalker314314:
Just now though - are you planning `./occ` command to change model?

matiasdelellis:
> Just now though - are you planning `./occ` command to change model?

It's already working.. Use `occ face:setup --model 3` :wink:

I would tell you 99% if it wasn't for the test that that fail.. :disappointed: 
There are also improvements to be made but it is already functional ..

matiasdelellis:
First statistics with real data.. :smile: 

```
[matias@nube nextcloud]$ sudo -u apache php occ face:setup --model 1 
The model 1 (Default) will be installed
The files of model 1 (Default) are already installed
This model was configured as default
[matias@nube nextcloud]$ sudo -u apache php occ face:stats -u matias
+--------+--------+-------+---------+
| User   | Images | Faces | Persons |
+--------+--------+-------+---------+
| matias | 2776   | 1792  | 882     |
+--------+--------+-------+---------+
[matias@nube nextcloud]$ sudo -u apache php occ face:setup --model 3
The model 3 (DlibHog) will be installed
The files of model 3 (DlibHog) are already installed
This model was configured as default
[matias@nube nextcloud]$ sudo -u apache php occ face:stats -u matias
+--------+--------+-------+---------+
| User   | Images | Faces | Persons |
+--------+--------+-------+---------+
| matias | 2776   | 1058  | 342     |
+--------+--------+-------+---------+
[matias@nube nextcloud]$ 
```

We could summarize it in that HOG found 59% of faces in my images, but clustered in 39% of persons.

CNN Model has 352 faces with a confidence less than 1 (Note that HOG don't report this). If we directly discard them (I am not going to do it), would be:
```
+--------+--------+-------+---------+
| User   | Images | Faces | Persons |
+--------+--------+-------+---------+
| matias | 2776   | 1440  | 530     |
+--------+--------+-------+---------+
```

So, HOG has 73,5 % of faces on 64,5% of persons..

The main groups are very similar, so without thinking much it could be assumed that the quality improved. I would not dare to assure it, but it is curious .. :sweat_smile:

So, the conclusions are still obvious.
 * CNN can find more faces that HOG
 * HOG can invent a face inside a lamp. (I found very few, and they were not clustered), but although less CNN too.
 * HOG practically does not require memory, and CNN needs many requirements.
 * HOG in general is faster that CNN, but with the current configuration by little.

Faces that are trash
imagen

matiasdelellis:
MMmmm..
Last analysis with with different sizes of images. (HOG in particular with larger size). So difficult to conclude, but still happy with the results.. :wink:

matiasdelellis:
I upload more images here to document a bit on the wiki .. :wink: 
imagen
imagen
imagen

matiasdelellis:
As a curiosity, I did the same analysis only with the same photo that you had originally used on 
* https://github.com/stalker314314/dlib-cnn-face-detection-analysis

Hog didn't find any faces in any size.. :sweat_smile: 
Cnn detected 42 faces on 1.501x1.126 images (1.690.000).. and using 6475.3 MB :wink:  

When we start the migration to pdlib we assume linearity for 1*1024 according to your analysis, and in the last release we change it to 4 (6475,3*1024/1690000 = 3,92).

Reviewing your original analysis again this difference is because pdlib uses [upsampling=1](https://github.com/goodspb/pdlib/blob/master/src/cnn_face_detection.cc#L54) by default, and it is consistent that we have to divide it by 4 according to your analysis..

So now I am happy that the results are similar. :smiley:

matiasdelellis:
https://github.com/matiasdelellis/facerecognition/wiki/Models
https://github.com/matiasdelellis/facerecognition/wiki/Comparative-analysis-of-the-models

matiasdelellis:
p.s. I will do the analysis again with upsampling=0
It is better that we create the correct image size directly, instead of reducing it a lot and that dlib increases it again.. :sweat_smile:

matiasdelellis:
Although travis seems broken[1], I think he is in a condition for a real review. :smile: 

NOTE [1]: https://travis-ci.org/github/matiasdelellis/facerecognition/builds/661326183?utm_source=github_status&utm_medium=notification here half failed with the same errors and half pass so I understand that it is a problem of travis .. I do not know.. :disappointed:

stalker314314:
I was following closely what you did, so no additional review is needed:D

PR #236, opened 2020-03-13
Move to new repo that use reprepro and test bionic

matiasdelellis:
:wink:


matiasdelellis:
Well.. The latter will fail because I can only compile it for 7.2

matiasdelellis:
Note that there is no reason to move to bionic beyond trying it and it is newer...

PR #239, opened 2020-03-27
Another class to handle the resizing of images more cleanly

matiasdelellis:
Just upload it to see scrutinizer report..

MM.. I am not convinced by the name "TempImage".. @stalker314314 What you think?


matiasdelellis:
Ok.. It remained as TempImage because it really is the only job it does. :sweat_smile: 

Refactoring the code of ImageProcessingTask, I could forget the methods of setFaces() and getFaces() methods that before involving some kind of context or another function.
Now the main code was a little bigger, but much cleaner, and I think understandable at a glance.

lib/Db/Face.php
```diff
@@ -138,18 +138,6 @@ public static function fromModel(int $image, array $faceFromModel): Face {
 		return $face;
 	}
 
-	/**
-	 * Helper method, to normalize face sizes back to original dimensions, based on ratio
-	 *
-	 * @param float $ratio Ratio of image resize
-	 */
```
stalker314314: Excellent for removing this (refactoring this whole normalize code! it looks better now!)

Issue #212, opened 2020-01-07
Support for external storage


Ang3er:
As already discussed in issue #26 a beta version was released for support of shared storage. In my case my nextcloud instance is running in a vm on my QNAP Nas. Thereby I mounted the storage as external storage with type local via an nfs mount. As nextcloud now thinks it is an external storage, the facerecognition is not analyzing any pictures in the external storages.

A screenshot to explan my mount option.
CUrrently it is not an option to change the mount option, as I need to access the storages via QNAP directly with afp. Therefore it is not an option to hold all data as a user local data and share it with the others, because than the data will be stored in nextcloud/data/user/files/...
Currently I can access the data directly via afp and the folder Bilder ;)

![image](https://user-images.githubusercontent.com/12126245/71893669-9bd62680-314c-11ea-92a2-51f7d95d5044.png)

Thanks in advance

matiasdelellis:
Hi @Ang3er 
The same logic used in #26 can be used for external storages.

https://github.com/matiasdelellis/facerecognition/blob/4a2683fab8a0da8d0bdfa1600a1fc29d5c7537f1/lib/Service/FileService.php#L192-L203

This is pending because I can't find a consistent way to detect them. We need something as simple as this.

https://github.com/matiasdelellis/facerecognition/blob/4a2683fab8a0da8d0bdfa1600a1fc29d5c7537f1/lib/Service/FileService.php#L178-L183

But I still don't see how to do it cleanly. And note that I cannot consider all cases individually. It should be something more or less generic.

Ang3er:
@matiasdelellis Perferct, didnt know that this request is still pending. Thanks for the nice development. Therefore I think we can close this issue

matiasdelellis:
Reopen this as reference.. :wink:

hairlesshobo:
Hey, great work on this app so far. Believe it or not, the ability to do in-house facial recognition is what encouraged me to setup my own NextCloud instance. With that said.. what is the latest news (if any) on support for external storage? All of my data is stored on external storage and without the ability to scan for faces there, it makes me sad ;).

Ang3er:
@matiasdelellis: May this be a reference point as the occ files:scan --all is enabled to check on external data storages 
https://help.nextcloud.com/t/automate-occ-filescan/35282/7

thanks in advance

matiasdelellis:
Hi @Ang3er @hairlesshobo @stalker314314 

[...]

PR #246, opened 2020-04-06
Experimental support of External Storage. Issue #212

matiasdelellis:
*Experimental support of External Storage. Issue #212 
* Fix that files shared within the same instance were considered native.
* Ignore not available storages.
* When it detects that a file is not enabled (When it deactivates the option of
* shared or external), it tries to clean up in the next run.

### DIFF
--- a/lib/Service/FileService.php
+++ b/lib/Service/FileService.php
@@ -33,6 +33,7 @@ use OCP\ITempManager;
 
 use OCP\Files\IHomeStorage;
 use OCP\Files\NotFoundException;
+use OCP\Files\StorageNotAvailableException;
 
 use OCA\Files_Sharing\External\Storage as SharingExternalStorage;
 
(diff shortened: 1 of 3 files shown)


lib/Service/FileService.php
```diff
@@ -184,17 +189,25 @@ public function setDescendantDetection(Folder $folder, bool $detection): bool {
...
+		return (($node->getMountPoint()->getMountType() === '') &&
+		        ($node->getStorage()->instanceOfStorage(IHomeStorage::class)));
+	}
+
+	/**
+	 * Returns if the file is inside a shared mount storage.
 	 */
 	public function isSharedFile(Node $node): bool {
-		return $node->getStorage()->instanceOfStorage(SharingExternalStorage::class);
+		return ($node->getMountPoint()->getMountType() === 'shared');
```
stalker314314, line 203: this is proper way to detect shared/local? I mean, with mountType and not with `instanceOfStorage`? If it is...why I used `instanceOfStorage` 🤔 ? Did I just overlooked this one, or this was added later?

matiasdelellis: What I discovered in recent tests, is that files shared by another user from the same instance are reported within HomeStorage. So,  it did not even run at this check. They were directly considered as local files. Therefore, to the local files I also verify that the mountype is also empty.

In my previous tests I only tested it with an external instance.. :disappointed:

stalker314314: I didn't know about `mountType` when I wrote it. Maybe I should just have used this from start:) Good you fixed it:D

matiasdelellis: MM.. I think it was a copypast made by me from another application.. :sweat_smile: 
But without enabling shared files (That still is experimental) everything worked fine. 

Well, I hope you can finally analyze your images!. :smile:

Issue #248, opened 2020-04-07
Issue - Seems that still don't configured the image area used for analysis


Floflobel:
### Actual behaviour

The command background_job return an error. 

### Server configuration

* **Operating system:** Ubuntu 18.04.4 LTS

* **Pdlib version:** Build manually

* **How is DLib installed:** Build manually
pdlib-min-test-suite works 

* **How is PDlib installed:** Bruild manually
pdlib-min-test-suite works

* **PHP version:** 7.2

* **Web server:** Apache 

* **Database:** Postgres

* **Nextcloud version:** 17.0.2

### Client configuration

* **Browser:** Firefox

* **Operating system:** Linux

#### Background task log with debug.
<details>
<summary>sudo -u apache php occ -vvv face:background_job</summary>

```
Insert your background log here
root@box:/var/www/nextcloud # sudo -u www-data php occ -vvv face:background_job
1/10 - Executing task CheckRequirementsTask (Check all requirements)
        Seems that still don't configured the image area used for analysis.
Please contact administrator to configure it in the admin panel to continue.
Fill an issue here if that doesn't help: https://github.com/matiasdelellis/facerecognition/issues
Task CheckRequirementsTask signalled we should not continue, bailing out
```
</details>

#### Web server error log
<details>
<summary>Web server error log</summary>

```
Nothing when run de background_job
```
</details>

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>

```
Nothing 
```
</details>

matiasdelellis:
Go to `[NEXTCLOUDURL]/settings/admin/facerecognition` :wink:

Floflobel:
![Screenshot from 2020-04-07 14-21-26](https://user-images.githubusercontent.com/13889750/78668514-1bee6280-78db-11ea-9fb1-1393f21c3624.png)
Yes, but I don't see any parameters that would solve the error ...

Floflobel:
Okay, I just hit the different bars and it's working. I looked all over the place but I didn't understand, it was so simple.

matiasdelellis:
Ohh.. It is the option of "Temporary files". Assign an area that seems logical to you, I recommend only 1024x768, which gives good results to start. This has a direct impact on memory consumption, therefore I recommend testing and increasing it later.

matiasdelellis:
I reopen as a reference, since I must improve the messages. 

Thanks!.

florian-bellencontre:
[...]

PR #249, opened 2020-04-08
Misc fixes v2

matiasdelellis:
* Add some documentation about how expect to install, test and use the application
* Try to clarify that the our commands are occ options and not executable within our application.
* Consider the memory configured in php as a dependency on the models. This will prevent the installation of models if you did not configure the memory. This is part of issue #248
* Fix a typo in the minimum area. Must be 640x480
* Also adds a minimum php memory requirement for HOG. 128 MB
* The maximum and minimum area for HOG is taken directly from SettingsService, since it is independent of the memory.
* Improves the description of errors and prints the links to documentation whenever it can.
* In the models table of --setup command indicate which one is enabled.

p.s: I'm probably going to upload small commits with other fixes, but I'm interested in discussing the readme ..

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,46 +36,61 @@ Finally the user can use the application in three ways
  3. In the side panel of the file application, a 'Persons' tab is added where
     you can see a list of your friends in the photo, and rename them.
 
-## Installation
+## Installation, configuration and usage
 
 #### Requirements
 
- * Nextcloud 14+
+ * Nextcloud 16+
  * [Dlib PHP bindings](https://github.com/goodspb/pdlib)
+ * 2GB of RAM
 
-Everything is AGPL or Creative Commons. :wink:
+#### Installation
 
-#### Manual installation on Ubuntu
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -36,46 +36,61 @@ Finally the user can use the application in three ways
...
 
-## Installation
+## Installation, configuration and usage
 
 #### Requirements
 
- * Nextcloud 14+
+ * Nextcloud 16+
  * [Dlib PHP bindings](https://github.com/goodspb/pdlib)
+ * 2GB of RAM
```
stalker314314, line 45: 2GB is not hard requirement anymore, not sure why you are adding this? Maybe add it, but with some disclaimer:)

matiasdelellis: In the last releases was established 2 GB of ram as a general requirement, with the excuse that the clustering of many photos can consume a lot of memory.

Then it requires 1GB as "memory_limits" to PHP for cnn models.

PR #250, opened 2020-04-09
Return more info when find an face... wink wink

matiasdelellis:
imagen

:see_no_evil: 

I don't want add a configuration on the server., and also don't really like the code in the FaceController, but there is nothing better.

This can be just something fun, or something useful for reporting bugs.

~~MMM.. I found this, that I found interesting: http://thatmikeflynn.com/egg.js/~~

~~What do you think that the browser requests it after typing the Konami code.?~~

I tried a little it, and it is very difficult to do without hardcoding a lot of code. So discarding it. The only viable thing is to add a configuration.


lib/Controller/FaceController.php
```diff
@@ -81,4 +122,82 @@ public function getThumb ($id, $size) {
 		return $resp;
 	}
 
+	/** @scrutinizer ignore-unused */
```
stalker314314, line 125: it is not unused anymore?:)

lib/Controller/FaceController.php
```diff
@@ -81,4 +122,82 @@ public function getThumb ($id, $size) {
...
+			return;
+		}
+
+		$eyesW = $eyesX2 - $eyesX1;
+		$eyesH = $eyesY2 - $eyesY1;
+
+		$eyesL = sqrt(pow($eyesW, 2) + pow($eyesH, 2));
+		$angle = rad2deg(atan(-$eyesH/$eyesW));
+
+		$glassesGd = imagecreatefrompng(\OC_App::getAppPath('facerecognition') . '/img/glasses.png');
```
stalker314314, line 165: should you mention somewhere is this image CC? do you need to mention attribution?

lib/Controller/SettingsController.php
```diff
@@ -197,6 +197,9 @@ public function setAppValue($type, $value) {
 			case SettingsService::SHOW_NOT_GROUPED_KEY:
 				$this->settingsService->setShowNotGrouped($value === 'true' ? true : false);
 				break;
+			case SettingsService::OFUSCATE_FACE_THUMBS_KEY:
```
stalker314314, line 200: `OFUSCATE_FACE_THUMBS_KEY` -> `O**B**FUSCATE_FACE_THUMBS_KEY`

`getOfuscateFaces` -> `getO**b**fuscateFaces`

stalker314314: actually, on a lot of places

lib/Db/FaceMapper.php
```diff
@@ -38,7 +38,7 @@ public function __construct(IDBConnection $db) {
 
 	public function find (int $faceId) {
 		$qb = $this->db->getQueryBuilder();
-		$qb->select('id', 'image', 'person', 'left', 'right', 'top', 'bottom', 'descriptor')
+		$qb->select('id', 'image', 'person', 'left', 'right', 'top', 'bottom', 'landmarks', 'descriptor', 'confidence')
```
stalker314314, line 41: #consider making landmarks under boolean argument (but I am not even sure we need this micro-optimization), feel free to ignore this:D

lib/Controller/FaceController.php
matiasdelellis: I was trying to avoid reports by the if (false). It never worked ..

lib/Controller/FaceController.php
matiasdelellis: Maybe .. But where? The application icons also not are licensed..

lib/Controller/SettingsController.php
matiasdelellis: Ohhh..

lib/Db/FaceMapper.php
matiasdelellis: I think I'm not going to do it, because it is a very punctual call, requesting a particular face, and I understand that it must return the entire entity.
If it is necessary in other calls surely I will do it, but 'find()' should only search by id without any restriction.

lib/Controller/FaceController.php
matiasdelellis: p.s: They are images originally with CC, that I modify to adapt, and they are still CC obviously.

Issue #244, opened 2020-03-31
`occ face:setup -m 3` doesn't indicate what deps are missing


strugee:
Note: omitted irrelevant sections of the issue template, like client configuration.

### Expected behaviour

When I run `occ face:setup -m 3` and receive an error about missing dependencies, I should be told what exactly is missing.

### Actual behaviour

I'm presented this with this generic error:

    % sudo -u www-data php -f /srv/http/nextcloud/occ face:setup -m 3
   You do not meet the dependencies to install the model 3 (DlibHog)

### Steps to reproduce

N/A basically

### Server configuration

* **Operating system:** Debian 10

* **Pdlib version:** 1.0

* **How is DLib installed:** from source (`make php-test` in the compatibility tool crashes with `std::bad_alloc`, presumably because I don't have enough memory, but I know it's installed okay because facial recognition has successfully run in the past - I just had to pass a parameter to `php` to jack up the memory limit)

* **How is PDlib installed:** from source: (the SaltStack equivalent of) `phpize && ./configure --prefix=/usr/local --enable-debug && make && sudo make install`

* **PHP version:** 7.3.14-1~deb10u1

* **Web server:** Apache 2

* **Database:** MySQL

* **Nextbutt version:** 18.0.3

#### Background task log with debug.

N/A

#### Web server error log

Nothing relevant.

#### Nextcloud log (data/nextcloud.log)

Nothing related is logged to the Nextcloud log.

matiasdelellis:
Hi,
Thank you for the report!. You are right, I must improve the messages .. :wink: 

Well, I understand that you had the application working, and you want to change to HOG for memory usage. No?.
The problem is that this model depends on a pdlib library that has not been released yet. :disappointed:

See: https://github.com/goodspb/pdlib/pull/20

We are seriously considering taking maintenance of this library, but in the meantime you should compile pdlib again from https://github.com/matiasdelellis/pdlib/tree/improve-hog

Sorry..

strugee:
No, don't apologize! Thank you so much for maintaining this application, I think it has a ton of potential :-)

I'll pin my SaltStack formulas to the tip of that branch and see if that fixes things. I'll also add a note to the wiki to that effect.

Thanks for the quick reply!

strugee:
Worked perfectly, and I see you've actually already edited the wiki. Thanks again :-)

I'll leave this ticket open for better error messages.

stalker314314:
[...]

PR #254, opened 2020-04-09
Use references to get a description of the problem. Issue #244

matiasdelellis:
Well.. This may seem out of the spirit of PHP, but I don't want to use exceptions everywhere ..


matiasdelellis:
Without many details, because I send it to read the documentation, which we must also improve.

PR #260, opened 2020-04-14
Use dlib_vector_length implemented on pdlib 1.0.2 

matiasdelellis:
See https://github.com/matiasdelellis/pdlib/pull/2

In a small benchmark with 1963 descriptors:

OLD php: 0.012276937961578 Sec.
New native: 0.0022540950775146 Sec.

This is 80% faster than our PHP class.

### DIFF
--- a/lib/BackgroundJob/Tasks/CreateClustersTask.php
+++ b/lib/BackgroundJob/Tasks/CreateClustersTask.php
@@ -232,19 +232,38 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 		$sensitivity = $this->settingsService->getSensitivity();
 		$min_confidence = $this->settingsService->getMinimumConfidence();
 		$edges = array();
-		for ($i = 0, $face_count1 = count($faces); $i < $face_count1; $i++) {
-			$face1 = $faces[$i];
-			if ($face1->confidence < $min_confidence) {
-				$edges[] = array($i, $i); // fixme: Should we create an single group? o just ignore.
-				continue;
-			}
-			for ($j = $i, $face_count2 = count($faces); $j < $face_count2; $j++) {
-				$face2 = $faces[$j];
-				// todo: can't this distance be a method in $face1->distance($face2)?
-				$distance = $euclidean->distance($face1->descriptor, $face2->descriptor);
 
-				if ($distance < $sensitivity) {
-					$edges[] = array($i, $j);
+		if (version_compare(phpversion('pdlib'), '1.0.2', '>=')) {
+			for ($i = 0, $face_count1 = count($faces); $i < $face_count1; $i++) {
+				$face1 = $faces[$i];
+				if ($face1->confidence < $min_confidence) {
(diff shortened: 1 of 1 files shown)


lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -232,19 +232,38 @@ private function getNewClusters(array $faces): array {
 		$sensitivity = $this->settingsService->getSensitivity();
 		$min_confidence = $this->settingsService->getMinimumConfidence();
 		$edges = array();
-		for ($i = 0, $face_count1 = count($faces); $i < $face_count1; $i++) {
-			$face1 = $faces[$i];
-			if ($face1->confidence < $min_confidence) {
-				$edges[] = array($i, $i); // fixme: Should we create an single group? o just ignore.
-				continue;
+
+		if (version_compare(phpversion('pdlib'), '1.0.2', '>=')) {
```
matiasdelellis, line 236: The benchmark would be to add `$startTime = microtime (true);`

lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -232,19 +232,38 @@ private function getNewClusters(array $faces): array {
...
+					continue;
+				}
+				for ($j = $i, $face_count2 = count($faces); $j < $face_count2; $j++) {
+					$face2 = $faces[$j];
+					$distance = dlib_vector_length($face1->descriptor, $face2->descriptor);
+
+					if ($distance < $sensitivity) {
+						$edges[] = array($i, $j);
+					}
+				}
```
matiasdelellis, line 250: ..and:

```
$stopTime = microtime (true);
print "\n\nNative length:" . ($stopTime-$startTime)/1000 . " Sec.\n\n"; 
```

lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -232,19 +232,38 @@ private function getNewClusters(array $faces): array {
...
 			}
-			for ($j = $i, $face_count2 = count($faces); $j < $face_count2; $j++) {
-				$face2 = $faces[$j];
-				// todo: can't this distance be a method in $face1->distance($face2)?
-				$distance = $euclidean->distance($face1->descriptor, $face2->descriptor);
+		} else {
```
matiasdelellis, line 252: And the same here to compare..
I ran it many times, and it's stable, but  more importantly, the resulting custers are the same ..

stalker314314:
Awesome! I was under impression that `dlib_chinese_whispers` was largest offender! This will speed up my calculation, it takes me 30 min and more to calculate clusters:o

matiasdelellis:
Keep in mind that this improves the obtaining of the edges.. But the same chinese_whisper is still used. :wink: 

Well, when you test it, please share the results in your original report.. https://github.com/matiasdelellis/facerecognition/issues/105

PR #265, opened 2020-04-22
Use paginated search. Fix part of issue #263

matiasdelellis:
This should improve search speed on files.

@stalker314314 I guess this affects you.

@DrRSatzteil If you can test it, I will thank you again.. :smile:


DrRSatzteil:
I will definitely try that. One question: is it ok to just clone and make the project again or is there anything I need to know when I'm updating like this?

DrRSatzteil:
I just gave it a try and it worked out fine. Unfortunately I rebuilt from the master branch 🙈 Will try again later...

matiasdelellis:
For this change from master, you just need to do `git checkout search-paged`
You shouldn't even do make again,

DrRSatzteil:
Thanks I tried from the correct branch now. However I do not see any improvement. The first query takes about a minute still. However when I search again the results are there more or less instantly. So apparently there is some cache in place somewhere. I also do not notice a difference UI-wise. It seemed to load page by page before the change already: below the list it states that there are 30 hits and after scrolling down it is corrected to 55 hits. This behaviour was the same in the master branch.

matiasdelellis:
Ohh.. Just to ensure something ..
Disable this application and search for anything in files.

DrRSatzteil:
With both fulltextsearch and facerecognition disabled I don't find any results.

matiasdelellis:
Don't serach persons, because you won't find it, but search for any file by name.. I want to make sure if there is any difference in performance.. :confused:

DrRSatzteil:
Oh well other files are found very quickly no matter if the app is enabled or not. What takes a while is below that where it says: detaching in different locations (an anderen Orten suchen - in German)

matiasdelellis:
> Oh well other files are found very quickly no matter if the app is enabled or not. What takes a while is below that where it says: detaching in different locations (an anderen Orten suchen - in German)

Ok .. Do you have any other application that indexes the files?. I ask you because for example Music also indexes without paging. :sweat_smile: Maybe this query is now fast, but other applications cause problems. I can continue to suggest deactivating applications, but already at this point it is disappointing. 
From my side I think that I cannot improve it anymore .. :disappointed:

DrRSatzteil:
Now I see what you're trying to find out. Well it seems that it is also searching for comments in files. This takes quite a while as well. I don't use Music however. Maybe I can test a bit tomorrow.

Issue #264, opened 2020-04-22
Wrong previews in sidebar when faces are incorrectly clustered in the same cluster


DrRSatzteil:
Sorry for raising the next issue but I discovered another unexpected behavior regarding the small preview faces in the sidebar.

### Expected behaviour
Lets use an actual example to allow for an easier explanation: I have a picture with my son, my daughter and one of the children of my neighborhood on it. All the faces were put into the same cluster. This is of course not correct but this is not my point here. I would now expect to see all the three faces in the sidebar with the same name next to it (because of the same cluster assignment). However...

### Actual behaviour

All three pictures in the sideline are identical, the same face is shown three times instead of the individual faces of each child. Unfortunately I cannot upload the picture or screenshots here but it should be easy to reproduce when you mess with the clustering settings.

### Steps to reproduce
1. Configure clustering badly so that all faces on a picture with three persons are put into the same cluster
2. Open the file in files and open the person sidebar
3. See the same face three times instead of the individual faces

### Server configuration

* **Operating system:**
Ubuntu 18.04

* **Pdlib version:**
php7.0-pdlib

* **How is DLib installed:** Make sure it is working correctly with this [tool](https://github.com/matiasdelellis/pdlib-min-test-suite)
As described in https://github.com/matiasdelellis/facerecognition/wiki/Installation "The Easy Way - Ubuntu Bionic"

* **How is PDlib installed:** Make sure it is working correctly with this [tool](https://github.com/matiasdelellis/pdlib-min-test-suite)
As described in https://github.com/matiasdelellis/facerecognition/wiki/Installation "The Easy Way - Ubuntu Bionic"

* **PHP version:**
7.2

* **Web server:**
Apache2

* **Database:**
MariaDB

* **Nextcloud version:**
18.0.3

matiasdelellis:
> I have a picture with my son, my daughter and one of the children of my neighborhood on it. All the faces were put into the same cluster.

About this unfortunately the models used have problems with childrens. This in the short term will not improve. Thanks for understanding.

> See the same face three times instead of the individual faces

Ok.. I could not check yet but it seems very feasible.. 

https://github.com/matiasdelellis/facerecognition/blob/master/lib/Controller/FileController.php#L98-L114

We are looking for faces of persons in files, a since it is the same person, it always returns the same face. We should look for persons of faces in files.

Thank you very much for reporting it. Soon news. :wink:

PR #266, opened 2020-04-22
Search mainly faces, rather than persons in file view. Fix issue #264

matiasdelellis:
@DrRSatzteil again you can test just with `git checkout fix-wrong-faces-file`


DrRSatzteil:
Looks perfect, thank you!

Issue #257, opened 2020-04-11
The persons side bar is not showed in photos


matiasdelellis:
### Describe the bug

The persons side bar is not showed in photos.

#### Steps to reproduce the behavior:

1. Go to files app
2. Click on some photo.
3.  Open the viewer, click "Open lateral pane".
4. Here see all tabs.. 😄
5. Go to photos app
6. Click on some photo.
7. Open the viewer, click "Open lateral pane".
8. Here see just "Chat" and "Shared" tabs. 😞

### Expected behavior
See persons tab in photos.

matiasdelellis:
Original report: https://github.com/nextcloud/photos/issues/243

PR #268, opened 2020-05-01
User new events to add the facerecognition tab

matiasdelellis:
Issue #257 

At this point I could only load our scripts with the sidebar.. still returns errors:

> TypeError: OCA.Files.DetailTabView is undefined

It is the same error of activity, comments, and version tabs.

I'm not sure if I should migrate to VUE, or is there an alternative.


lib/Listener/LoadSidebarListener.php
```diff
@@ -0,0 +1,48 @@
...
+	public function handle(Event $event): void {
+		if (!($event instanceof LoadSidebar)) {
+			return;
+		}
+
+		\OCP\Util::addScript('facerecognition', 'vendor/lozad');
+		\OCP\Util::addScript('facerecognition', 'fr-dialogs');
+		\OCP\Util::addScript('facerecognition', 'files-tabview');
+		\OCP\Util::addScript('facerecognition', 'files-plugin');
+		\OCP\Util::addStyle('facerecognition', 'files-tabview');
```
skjnldsv, line 45:
```suggestion
		Util::addScript('facerecognition', 'vendor/lozad');
		Util::addScript('facerecognition', 'fr-dialogs');
		Util::addScript('facerecognition', 'files-tabview');
		Util::addScript('facerecognition', 'files-plugin');
		Util::addStyle('facerecognition', 'files-tabview');
```

matiasdelellis: Ok.. @skjnldsv
Thanks for all the answers .. I will try to migrate to vue.. :wink:

skjnldsv: I might add the proper global standard for 20 though :see_no_evil:

matiasdelellis: imagen

I dont have any idea what I'm doing .. but at least it shows something .. hahah

matiasdelellis:
imagen

Wow.. Is working!. :smile:

lib/AppInfo/Application.php
```diff
@@ -47,6 +51,7 @@ public function __construct(array $urlParams = []) {
 
 		$this->connectWatcher();
 		$this->connectSearch();
+		$this->addServiceListenders();
```
stalker314314, line 54: Listenders->Listeners maybe

matiasdelellis:
@stalker314314 
I'm going to merge this change.. Beyond the increase in size, (From nothing to 145kb :sweat_smile:), the problem is that this depends on NC18.
My idea was to abandon NC16, but not NC17. Do you have a problem with this? :disappointed:

matiasdelellis:
> Congratulation Matias, you really rock! (I really don't have any other meaningful comment to add:D)

haha. Thanks, but all the background task is thanks to you!. :smile:

Issue #158, opened 2019-08-07
Feature: Able to move between groups or remove recognized faces. 


aleksandrmetik:
Hello
Could you please implement action to move recognized faces between groups and/or remove from group? Thanks

stalker314314:
This is similar to #134 . This will require another "layer" on top of basic clustering (manual layer) and we are not sure how to do it correctly. Manual actions (join clusters, move/remove face to group) needs to be preserved once new clustering is executed. There is no plans to do it soon

matiasdelellis:
Hi @aleksandrmetik,
As Branko says, It is an expected feature of an application like this, and we have to do it right.

So, this will be for a second stage.. With enough external testing we must first ensure the stability of the clusters ... and seriously think about how to implement it.

DerOetzi:
Hi @matiasdelellis,

for me a quick win would be to have the possibility in UI to untag (remove) single and complete groups of recognized faces completly from database, for two reasons:

- When I reduce sensitivity the clustering works very well for all of my friends, but there are always a few wrong faces in the cluster but mainly "unknown" persons from background, which I never want to be tagged. 
- For doing lot of pictures at the performance of my band there are lot of pictures of audience which are recognized as well and I want to remove them as well. Because they slow down clustering process unnecessarily. The initial scan process of my round about 20000 pictures is finished for about 70% and there are already 25.000 faces recognized. I restart background scan process every 60 minutes with a timeout of 58 minutes. Reclustering takes already about 10 minutes of each run. 

So I would request two feature:
- Remove recognized faces single and bulk.
- Requeue single picture for scanning, if I remove face accidentaly

matiasdelellis:
Hi @DerOetzi 

Thank you very much for testing the application. :wink:

What I can tell you is that it is still under evaluation.  
Now we have a simple database where to store the face encodings, and the clustering algorithm that gave us the best result keeping the application simple.

I am 95% sure that the wrong people in your main groups are unknown. But it doesn't mix two persons who are really important to you. If so, it is within current expectations.

I clarify and ask this, because one thing is remove a person that you do not care (A small hack), and another very different problem is to separate a group into two people (An important rewrite.)

Again, of course we have to improve, but I want to advance at the right step, and therefore I am thinking it through. That is, I don't want to do a little hack to add this feature, and then have to undo it because I found a better way or it causes problems.

This will take some time.. Sorry.

[...]

PR #280, opened 2020-06-05
[Draft] Delete single recognized face

DerOetzi:
This is a draft for my feature request on #158 to removed single tagged faces from database.

- [x] RestAPI Endpoint DELETE /face/{id}
- [ ] Integration UI Personal Site
- [x] Integration UI Filebrowser PersonsTab

Signed-off-by: Johannes DerOetzi Ott <info@johannes-ott.net>

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -31,6 +31,12 @@ return ['routes' =>
 		'url'  => '/face/{id}/thumb/{size}',
 		'verb' => 'GET'
 	],
+	// Delete a single face
+	[
+		'name' => 'face#deleteFace',
+		'url'  => '/face/{id}',
+		'verb' => 'DELETE'
+	],
 	// Get persons from path
 	[
 		'name' => 'file#getPersonsFromPath',
--- a/css/facerecognition.scss
+++ b/css/facerecognition.scss
@@ -23,15 +23,26 @@
 }
 
 .face-preview {
-	background-color: rgba(210, 210, 210, .75);
+	background-color: rgba(210, 210, 210, 0.75);
 	border-radius: 3px;
 	margin: 2px;
 	height: 50px;
 	width: 50px;
+
+	.icon {
+		display: none;
+		height: 17px;
+		width: 17px;
+		cursor: pointer;
+	}
 }
 
 .face-preview:hover {
-	opacity: .5;
+	opacity: 0.5;
+
+	.icon {
+		display: inline-block;
+	}
 }
 
 /*
@@ -44,7 +55,7 @@
 }
 
 .face-preview-dialog {
-	background-color: rgba(210, 210, 210, .75);
+	background-color: rgba(210, 210, 210, 0.75);
 	border-radius: 25px;
 	height: 50px;
 	width: 50px;
--- a/js/fr-dialogs.js
+++ b/js/fr-dialogs.js
@@ -87,6 +87,66 @@ const FrDialogs = {
 			input.select();
 		});
 	},
+
+	deleteFace: function (thumbUrl, callback) {
+		return $.when(this._getMessageTemplate()).then(function ($tmpl) {
+			var dialogName = 'fr-dialog-content';
+			var dialogId = '#' + dialogName;
+			var $dlg = $tmpl.octemplate({
+				dialog_name: dialogName,
+				title: t('facerecognition', 'Remove tagged face'),
+				message: t('facerecognition', 'Really remove recognized face?'),
+				type: 'notice'
+			});
+			var div = $('<div/>').attr('style', 'display:flex; align-items: center');
+			var thumb = $('<img class="face-preview-dialog" src="' + thumbUrl + '" width="50" height="50"/>');
+
+			div.append(thumb);
+			$dlg.append(div);
+
+			$('body').append($dlg);
+
+			// wrap callback in _.once():
+			// only call callback once and not twice (button handler and close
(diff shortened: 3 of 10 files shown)


matiasdelellis:
Wow @DerOetzi 
I see you took the work seriously. :grimacing: 
I'm not sure about this yet, but I'm going to guide you to do it right. :wink: 

I explain here why I said a "small hack". :wink: 

>  Remove recognized faces single and bulk.

It is tempting to delete the face as you are doing, but keep in mind that a face can be an entire cluster. You can leave a person wihtout faces, which breaks the integrity of the database. You must control this situation.

You can take ideas to improve in code of delete files. 
* https://github.com/matiasdelellis/facerecognition/blob/master/lib/Watcher.php#L239

> Requeue single picture for scanning, if I remove face accidentaly.

I showed you the code of deleting a photo, but it is not the correct approach. Obtaining this information is very expensive, and therefore you should not delete it unnecessarily. You must find a way to invalidate it.

Maybe add a `state` column to the faces table, where `0` is `ok`, and `1` is 'unknown person`, etc. (I didn't think much about names or values!) and adjust the queries to this parameter. 

@stalker314314 if you want to comment welcome.

> Requeue single picture for scanning

This in particular is something I would like, (Especially to use when improve the server and you can use more processing power), but it is not done even in the medium term. :sweat_smile:

matiasdelellis:
>  Integration UI Personal Site

For now don't worry about this. This view was developed to evaluate the quality of the clusters, but the main view should be the files and photos applications.

When we develop the people view in photos, this view will be removed. :wink: 
If the user wants to delete/invalidate/etc a face that opens the image and does it in the sidebar.

I already said it many times. But we not pretend to be better than google .. :sweat_smile: 

imagen

They by default show the most important faces, and allow you to add or remove another faces. but never eliminate this.. :wink:

stalker314314:
Hi @DerOetzi - nice code and thanks for doing this! However, I don't think this is best approach (IMHO)! Output of face recognition and chinese whispers (clustering logic) is rather unknown and unpredictable and we tend to try to keep it "as-is" and only operate on higher levels. I would rather see additional table, on top of current faces, that can be consulted to tell if face should be removed, moved to another person... If this looks like overkill, then @matiasdelellis approach with new `state` column also should work. In other words, I would rather let user fiddle with "presentation" layer, than with output of ML models. I think @matiasdelellis even fiddled with adding something like `state` column in https://github.com/matiasdelellis/facerecognition/pull/262, maybe you should look at that. I would not mind adding new state column in `Faces` (but we will still need additional table, or another column in `Faces` to "move" face from one person/cluster to another). Current solution, while I admire your positive attitude, is maybe hackish? Hack that work, I admit!:), but looks hackish to me. What do toy think @DerOetzi on this approach (with new table, or column)?

Another, unrelated note: what if pictures changes slightly and we need to re-run face recognition - that would be completely other face? Or user wipes everything and recreates from scratch? User will again have to deal with face removal. Can we be better than this? Can we, instead of removing `face_id`, remove some other identifier, let's say - `(image_id, left, right, top, bottom)`? Have a table `removed_faces` and keep those? But it will not help if image changes, we would still have to reset this too (as we don't know if image filename is the same and it is completely different picture)

DerOetzi:
Thanks to all of your feedback. I didn't thought about reaching the goal with first draft. So don't worry about my work. For me as an agile software architect it's a lot of fun to do some experiments to get closer to a possible solution by discussing the results of the experiment. I always do things like try out and fail fast and try something else.

To me UI/UX is very important as well as clean code I think I have lot of experience to those especially UX and maybe can give some input.

First of all I'm a fan of the idea of managing table(s), but like I wrote in #158 in my opinion especially the removal of garbage faces should not only be taken into account for viewing reasons but although for clean up of "data garbage" for performance reasons. Keeping the database smaller, could for example help the clustering process and although performance of simple queries. So I'm still not 100 percent convinced whether leaving ML-output-data "as-is" is really the best approach. But as I said above I can think about a new experiment about this approach.

> It is tempting to delete the face as you are doing, but keep in mind that a face can be an entire cluster. You can leave a person without faces, which breaks the integrity of the database. You must control this situation.

I already have come to this point, that there is the need of clean up persons table as well, but I was not sure. There is a TODO in my code in FaceMapper taking account of this question, which you answered now.

> Integration UI Personal Site
> 
> For now don't worry about this. This view was developed to evaluate the quality of the clusters, but the main view should be the files and photos applications.
> 
> When we develop the people view in photos, this view will be removed. 😉
> If the user wants to delete/invalidate/etc a face that opens the image and does it in the sidebar.

Removing this view in my opinion would be a step backwards in UX of the application. Let's take my collection of pictures. There are round about 20000 pictures at the moment with 28000 faces recognized, per shooting there about 100-200 new Pictures and because of lot of audience round about 100-300 new faces round about 80% unknown people. Approx 30 new  correct clustered persons and 20 mixed clustered persons  which are all unknown, 10-20 added to existing persons, and maybe some wrong added. Managing of the new 300 faces is much easier on a overview page like this view. For example removing a lot of them already by removing the 50 clusters is much faster, then clicking through  200 pictures and remove face by face. Although getting overview of persons and there faces is already very good at the moment, and could be improved with different  sorting, filtering and some more list handling options. I would find it a shame to loose this entry point to recognized faces. In my opinion this entry point is as important as more possibilities at single picture editor.

matiasdelellis:
Hi, thank you so much for your contribution. It's implemented with detach

Issue #284, opened 2020-06-18
 Exception: Call to a member function meetDependencies() on null


HollowedEmpire:
### Steps to reproduce
1. Go to Nextcloud's Settings
2. Click on Face Recognition under the Administration section

### Expected behaviour
When going to Face Recognition under the Admin section of Nextcloud's Settings page, I'm expecting some sort of page with settings.

### Actual behaviour
When going to Face Recognition under the Admin section of NextCloud's Settings page, I am directed to an internal error page and an exception is logged:

Exception: Call to a member function meetDependencies() on null

## Server configuration detail

**Operating system:** Linux 4.19.118-v7l+ #1311 SMP Mon Apr 27 14:26:42 BST 2020 armv7l

**Webserver:** Apache/2.4.38 (Raspbian) (fpm-fcgi)

**Database:** mysql 10.3.22

**PHP version:** 

7.3.14-1~deb10u1
Modules loaded: Core, date, libxml, openssl, pcre, zlib, filter, hash, Reflection, SPL, session, sodium, standard, cgi-fcgi, pdlib, mysqlnd, PDO, xml, apcu, bz2, calendar, ctype, curl, dom, mbstring, fileinfo, ftp, gd, gettext, gmp, iconv, igbinary, imagick, intl, json, exif, mysqli, pdo_mysql, pdo_pgsql, pdo_sqlite, pgsql, apc, posix, pspell, readline, redis, shmop, SimpleXML, sockets, sqlite3, sysvmsg, sysvsem, sysvshm, tokenizer, wddx, wikidiff2, xmlreader, xmlwriter, xsl, zip, Phar, Zend OPcache

**Pdlib version:**
PHP won't give a version number for pdlib, and readme from source does not have version either. It was freshly compiled as of today.

**How is DLib installed:** 
Via following the manual install steps from wiki - the "hard way"
https://github.com/matiasdelellis/facerecognition/wiki/Installation#hard-way
Testing detects few faces but it passes without error. (0 in frist image, 4 in second)

**How is PDlib installed:** 
Via following the manual install steps from wiki - the "hard way"
https://github.com/matiasdelellis/facerecognition/wiki/Installation#hard-way
Testing detects few faces but it passes without error. (0 in frist image, 4 in second)

**Nextcloud version:** 18.0.6 - 18.0.6.0

**Updated from an older Nextcloud/ownCloud or fresh install:** Updated from previous versions

**Where did you install Nextcloud from:** Download

[...]

PR #285, opened 2020-06-19
Fix Admin page when not model installed. Issue #284

matiasdelellis:


HollowedEmpire:
This resolved the issue for me and I was able to get to the admin screen without issues.

PR #287, opened 2020-06-29
Added Dockerfile

cab404:
It's really (took me 3 hours) painful to install deps for this plugin, so here's a Dockerfile.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM nextcloud
+
+RUN apt-get update
+RUN apt-get install -y \
+  git libopenblas-dev \
+  liblapack-dev libx11-dev \
+  pkg-config cmake libbz2-dev
+
+
+# install dlib
+WORKDIR /
+RUN git clone https://github.com/davisking/dlib.git
+RUN mkdir /dlib/dlib/build
+WORKDIR /dlib/dlib/build
+RUN cmake -DBUILD_SHARED_LIBS=ON ..
+RUN make -j$(nproc)
+RUN make install
+RUN rm -rf /dlib
+
+# install pdlib
+WORKDIR /usr/src/php/ext/
+RUN git clone https://github.com/goodspb/pdlib.git
+WORKDIR /usr/src/php/ext/pdlib
+RUN phpize
+RUN ./configure --enable-debug
+RUN make -j$(nproc)
+
+# Installing extensions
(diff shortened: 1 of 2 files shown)


cab404:
Dockerfile can definitely be improved, and separated into a multistage build for less clutter.

mgafner:
Thank you very much for the Dockerfile! I just used it and it works for my setup :-)

cab404:
> Thank you very much for the Dockerfile! I just used it and it works for my setup :-)

Nice)
Then I am fixing some minor issues with it, and it will be ready for merging.

matiasdelellis:
Hi @cab404 
Thank you very much for trying to help.. :smile: 

There was a lot of discussion about it, but we understand that here is not the place to add docker support. For example, our first docker users use alpine, so your dockerfile not support it..

Since we cannot add support to everyone, our decision was to document it on a wiki (Although now I see that it is not well indicated anywhere :disappointed: ).

* https://github.com/matiasdelellis/pdlib/wiki/Docker

If you don't want to help with the documentation there, more that welcome. :smile:
But I won't add single case support here. Sorry.. :disappointed: 

On the other hand, a specific example like yours would be well seen in the [nextcloud/docker](https://github.com/nextcloud/docker/tree/master/.examples) examples repository. I think it is a very good example on how to expand the official docker. :smiley:

matiasdelellis:
Add link on wiki:
https://github.com/matiasdelellis/facerecognition/wiki/Installation#docker-way

Thanks.

cab404:
> Add link on wiki:
> https://github.com/matiasdelellis/facerecognition/wiki/Installation#docker-way
> 
> Thanks.

No problems) 

I just want people to spend less time making Dockerfiles by themselves, and section you added mostly fixes that — although I would've added at least a mention about Docker in README.md in the installation section (cause first thing I do in long text files is keyword search).

joao-p-marques:
Hi @matiasdelellis 

Thanks a lot for all the work on this. I understand that maintaining a docker image might not be in you direct interest, but would you potentially accept a Dockerfile with both versions (alpine a debian based)? 
A local build can take forever, especially when talking about a little homeserver, or not work at all because of RAM limitations. 

An automated multi-arch build is not that hard to integrate with Github Actions and I would be happy to contribute it, if you like the idea :smiley:

Issue #290, opened 2020-07-08
Can you actively trigger analysis when a file changes (such as uploading a new file)?


medivh-jay:
Can you actively trigger analysis when a file changes (such as uploading a new file)?

Thank you very much

matiasdelellis:
HI @medivh-jay 
The new images are queued and analyzed correctly, but are not incorporated into the facial clustering until there are 10 new faces.. This is to avoid re-clusterig continuously. But it is certainly something to improve.

medivh-jay:
thank you very much. another problem is that there seems to be no Person tab in the left panel of the File or the left panel of the album

matiasdelellis:
Hi @medivh-jay 

> The new images are queued and analyzed correctly, but are not incorporated into the facial clustering until there are 10 new faces.. This is to avoid re-clusterig continuously. But it is certainly something to improve.

Ok. In #291 check the conditions of the recreation of the clusters.
As I said there was a minimum of 10 new faces (Beyond that it is mainly a cleaning, I increased it to 25 in #291) to recreate these, but there is also a 2 hour max condition. So if there is at least one face that was analyzed more than two hours ago, the clusters will be recreated to incorporate it. This never worked for you?

> thank you very much. another problem is that there seems to be no Person tab in the left panel of the File or the left panel of the album

This is strange, can you give details about your installation, browser and please check the logs? :confused:

medivh-jay:
Hi @matiasdelellis 

> So if there is at least one face that was analyzed more than two hours ago, the clusters will be recreated to incorporate it. This never worked for you? 

It's worked, thank you for your reply

> This is strange, can you give details about your installation, browser and please check the logs? 😕

My nextcloud server version is 18.0.6,
and my browser is Chrome latest version
and my PHP version is 7.4.6
Here are all the plug-ins I installed
Enabled:
  - accessibility: 1.4.0
  - activity: 2.11.0
  - admin_audit: 1.8.0
  - apporder: 0.10.0
  - audioplayer_editor: 0.2.2
  - breezedark: 18.0.20
  - bruteforcesettings: 1.6.0
  - calendar: 2.0.3
  - cloud_federation_api: 1.1.0
  - comments: 1.8.0
  - contacts: 3.3.0
  - dav: 1.14.0
  - documentserver_community: 0.1.7
  - drawio: 0.9.6
  - external: 3.5.0
  - facerecognition: 0.6.1
  - federatedfilesharing: 1.8.0
[...]

PR #291, opened 2020-07-08
Small rewriting of cluster creation conditioning.

matiasdelellis:
The idea was to improve the initial conditions of recreation, but it became a little rewrite to improve the reading of the code.

It still has a bug, that I noticed a long time ago but never solved it.. and the issue #290 made me remember. :sweat_smile:
Basically, if you have more than 100 photos, analyzed or not, the clusters will be created. :disappointed:

### DIFF
--- a/lib/BackgroundJob/Tasks/CreateClustersTask.php
+++ b/lib/BackgroundJob/Tasks/CreateClustersTask.php
@@ -104,87 +104,53 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 	}
 
 	private function createClusterIfNeeded(string $userId) {
-		// Check that we processed enough images to start creating clusters
-		//
 		$modelId = $this->settingsService->getCurrentFaceModel();
-
-		$hasPersons = $this->personMapper->countPersons($userId, $modelId) > 0;
 
 		// Depending on whether we already have clusters, decide if we should create/recreate them.
 		//
+		$hasPersons = $this->personMapper->countPersons($userId, $modelId) > 0;
 		if ($hasPersons) {
-			// OK, we already got some persons. We now need to evaluate whether we want to recreate clusters.
-			// We want to recreate clusters/persons if:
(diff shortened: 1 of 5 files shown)


lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -104,87 +104,53 @@ public function execute(FaceRecognitionContext $context) {
...
-				$percentImagesProcessed = $imageProcessed / floatval($imageCount);
+			$forceTestCreation = $this->settingsService->_getForceCreateClusters($userId);
+			$needCreate = $this->needCreateFirstTime($userId, $modelId);
+
+			if ($forceTestCreation) {
+				$this->logInfo('Force the creation of clusters for testing');
 			}
-			$facesCount = $this->faceMapper->countFaces($userId, $modelId);
-			// todo: get rid of magic numbers (move to config)
-			if (!$forceCreation && ($facesCount < 1000) && ($imageCount < 100) && ($percentImagesProcessed < 0.95)) {
```
matiasdelellis, line 177: @stalker314314 
Do you remember what your original intention here?
As it was (and still in this PR), having 100 photos, analyzed or not, always creates the initial clusters.

lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -210,7 +176,83 @@ private function createClusterIfNeeded(string $userId) {
...
+		// todo: get rid of magic numbers (move to config)
+		$facesCount = $this->faceMapper->countFaces($userId, $modelId);
+		if ($facesCount > 1000)
+			return true;
+
+		if ($imageCount > 100)
+			return true;
+
+		$percentImagesProcessed = $imageProcessed / floatval($imageCount);
+		if ($percentImagesProcessed < 0.95)
```
matiasdelellis, line 252: D'Oh!. jaja

PR #299, opened 2020-07-17
Add an option to not cluster smaller faces.

matiasdelellis:
Now in my database, i have so many faces of just 30 pixels. Most of all, poor
quality WhatsApp photos, but since I increase them to 2Mpx to analyze, the
process finds an lot of that faces.

I like to see them, but nobody can believe that with 30px you can reliably
compare something. So I just don't group them.

As you can imagine, this is inspired by the last (super promising and failed :sweat_smile:) branch... 

Well, I read a lot of documentation, and the recognition model is indeed trained for 150px faces.
Anything smaller than that can be problematic, however the default value is set to 125. Just to have some margin..

However, it is worth noting that this value is taken from the real image.. So, if you have a 150 pixel face, in a very large image, that eventually shrinks during analysis, the descriptor is obtained from an image of less than 150px, which will not be reliable. :disappointed: 

To know if it is reliable, we should save the ratio and calculate the size used on analysis. But it is out of this PR.
Also, we could crop the image over the rectangle of the face and analyze the original, but I would rather avoid it.  Remember that? https://github.com/matiasdelellis/facerecognition/pull/142 :see_no_evil: 

Well, as I said in the commit, a 30px face is not reliable, and at least I avoid these cases..

### DIFF
--- a/lib/BackgroundJob/Tasks/CreateClustersTask.php
+++ b/lib/BackgroundJob/Tasks/CreateClustersTask.php
@@ -278,14 +278,14 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 			for ($i = 0; $i < $faces_count; $i++) {
 				$face1 = $faces[$i];
 				if (($face1->confidence < $min_confidence) ||
-				    ($face1->height() < $min_face_size)) {
+				    (max($face1->height(), $face1->width()) < $min_face_size)) {
 					$edges[] = array($i, $i);
 					continue;
 				}
 				for ($j = $i; $j < $faces_count; $j++) {
 					$face2 = $faces[$j];
 					if (($face2->confidence < $min_confidence) ||
-					    ($face2->height() < $min_face_size)) {
+					    (max($face2->height(), $face2->width()) < $min_face_size)) {
 						continue;
 					}
 					$distance = dlib_vector_length($face1->descriptor, $face2->descriptor);
@@ -300,14 +300,14 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 			for ($i = 0; $i < $faces_count; $i++) {
 				$face1 = $faces[$i];
 				if (($face1->confidence < $min_confidence) ||
-				    ($face1->height() < $min_face_size)) {
+				    (max($face1->height(), $face1->width()) < $min_face_size)) {
 					$edges[] = array($i, $i);
 					continue;
 				}
 				for ($j = $i; $j < $faces_count; $j++) {
 					$face2 = $faces[$j];
 					if (($face2->confidence < $min_confidence) ||
-					    ($face2->height() < $min_face_size)) {
+					    (max($face2->height(), $face2->width()) < $min_face_size)) {
 						continue;
 					}
 					// todo: can't this distance be a method in $face1->distance($face2)?
--- a/lib/BackgroundJob/Tasks/CreateClustersTask.php
+++ b/lib/BackgroundJob/Tasks/CreateClustersTask.php
@@ -265,22 +265,27 @@ class CreateClustersTask extends FaceRecognitionBackgroundTask {
 	}
 
 	private function getNewClusters(array $faces): array {
-		// Create edges for chinese whispers
-		$euclidean = new Euclidean();
+		// Clustering parameters
 		$sensitivity = $this->settingsService->getSensitivity();
 		$min_confidence = $this->settingsService->getMinimumConfidence();
+		$min_face_size = $this->settingsService->getMinimumFaceSize();
+
+		// Create edges for chinese whispers
 		$edges = array();
 
 		if (version_compare(phpversion('pdlib'), '1.0.2', '>=')) {
-			for ($i = 0, $face_count1 = count($faces); $i < $face_count1; $i++) {
+			$faces_count = count($faces);
+			for ($i = 0; $i < $faces_count; $i++) {
 				$face1 = $faces[$i];
-				if ($face1->confidence < $min_confidence) {
+				if (($face1->confidence < $min_confidence) ||
+				    ($face1->height() < $min_face_size)) {
 					$edges[] = array($i, $i);
 					continue;
 				}
-				for ($j = $i, $face_count2 = count($faces); $j < $face_count2; $j++) {
+				for ($j = $i; $j < $faces_count; $j++) {
 					$face2 = $faces[$j];
-					if ($face2->confidence < $min_confidence) {
+					if (($face2->confidence < $min_confidence) ||
+					    ($face2->height() < $min_face_size)) {
 						continue;
 					}
 					$distance = dlib_vector_length($face1->descriptor, $face2->descriptor);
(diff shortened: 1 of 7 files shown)


lib/Db/FaceMapper.php
```diff
@@ -131,7 +131,7 @@ public function getOldestCreatedFaceWithoutPerson(string $userId, int $model) {
 
 	public function getFaces(string $userId, $model): array {
 		$qb = $this->db->getQueryBuilder();
-		$qb->select('f.id', 'f.person', 'f.confidence', 'f.descriptor')
+		$qb->select('f.id', 'f.person', 'f.bottom', 'f.top', 'f.confidence', 'f.descriptor')
```
matiasdelellis, line 134: TODO: We could create two methods, one for the clustering stage, and another for the fronted code to optimize...

lib/Helper/TempImage.php
```diff
@@ -109,9 +109,13 @@ public function clean() {
 	 *
 	 */
 	private function prepareImage() {
-		$this->loadFromFile($this->imagePath);
-		$this->fixOrientation();
```
matiasdelellis, line 113: D'Oh.. This was a error...

matiasdelellis:
p.s:

**Master:**
```
	Clusters already exist, but there was some change that requires recreating the clusters
	2045 faces found for clustering
	1103 persons found after clustering
```

**This PR:**
```
	Clusters already exist, but there was some change that requires recreating the clusters
	2045 faces found for clustering
	1236 persons found after clustering
```

Basically 133 faces (6.5% of mi faces) that are not clustered now. It may seem like a lot, but considering that the main clusters practically do not change (355 faces in Master and 336 whit this PR: 5.4% less), i want to believe that clusters could only improve. :sweat_smile:

In short, I want to have the main clusters with high quality, and then if there are thousands of secondary clusters, I don't worry, but they will probably also improve ..

templates/settings/admin.php
```diff
@@ -55,6 +55,21 @@
 			<a id="save-min-confidence" class="icon-align icon-edit" style="display: none;" title="<?php p($l->t('Save'));?>" href="#"></a>
 		</p>
 		<br>
+		<h3>
+			<?php p($l->t('Minimum face size'));?>
+		</h3>
+		<p class="settings-hint"><?php p($l->t('Very small faces may be discovered during analysis, however they can be very diffuse to compare reliably. The smaller faces will not be clustered, but you will still be able to identify them.'));?>
```
stalker314314, line 61: > but you will still be able to identify them.

What this means? If I set min face to 50x50px, they will not be in clustering, right? How can I "identify"them? If I open image with them, I will see them, but no person will be assigned to them?

Asking as this might be confusing to users (or maybe it is confusing to me only:D)

lib/BackgroundJob/Tasks/CreateClustersTask.php
```diff
@@ -290,20 +295,23 @@ private function getNewClusters(array $faces): array {
...
 			}
 		} else {
-			for ($i = 0, $face_count1 = count($faces); $i < $face_count1; $i++) {
+			$euclidean = new Euclidean();
+			$faces_count = count($faces);
+			for ($i = 0; $i < $faces_count; $i++) {
 				$face1 = $faces[$i];
-				if ($face1->confidence < $min_confidence) {
+				if (($face1->confidence < $min_confidence) ||
+				    ($face1->height() < $min_face_size)) {
```
stalker314314, line 303: why do you compare height only? Is our model always returning height>width for found faces? In theory, you probably want here `max($face1->height(), $face1->width()) < $min_face_size`? What about orientation, if image is landscape, is width>height for found faces, or it is already accounted (rotated)?

lib/Service/SettingsService.php
```diff
@@ -63,6 +63,12 @@ class SettingsService {
 	const DEFAULT_MINIMUM_CONFIDENCE = '0.99';
 	const MAXIMUM_MINIMUM_CONFIDENCE = '1.1';
 
+	/** Minimum face size used to try to clustring faces */
+	const MINIMUM_FACE_SIZE_KEY = 'min_face_size';
+	const MINIMUM_MINIMUM_FACE_SIZE = '0';
+	const DEFAULT_MINIMUM_FACE_SIZE = '125';
```
stalker314314, line 69: Why do you do this, because small faces tend to give wrong result/bad 128D descriptors? Or to remove non-important stuff/features from image? If former, it makes sense to have this as absolute value (in pixels). But, if latter, maybe you should define `min_face_`size` as percentage from image size? Just saying, you probably though of this more than me:D

templates/settings/admin.php
matiasdelellis: > If I open image with them, I will see them, but no person will be assigned to them?

Now the face will now be clustered as a single photo person. 

> I will see them, but no person will be assigned to them?

This would be ideal. I always get annoyed by the `New Person ##` of the clusters.  But now it doesn't change.

I was thinking, make that the name admit NULL, and the user assigns his name directly.
* If the cluster has a name assigned, never re-cluster their faces.
* If the user assigns a NULL cluster to existing name,link to original cluster.
* So, re clustering just the faces on unnamed clusters.

But to do something like this, you have to have good clusters first, which is what I'm trying to improve these days.

matiasdelellis: p.s: You can suggest the message.. :wink:

lib/BackgroundJob/Tasks/CreateClustersTask.php
matiasdelellis: Just to optimize (Although probably exaggerated ..), and because I assume that the faces are always vertical perhaps change it.
But in general the rectangles are quite square, and if not, it is another poor quality alarm. :thinking:

lib/Service/SettingsService.php
matiasdelellis: > Why do you do this, because small faces tend to give wrong result/bad 128D descriptors?

Yes. I use 125 instead of 150 because I assume 125 are still good faces. But the famous 99.34% quality of the model is with 150px faces and good quality.

Note that there are many photos, which seems good at a glance, but if you only see the detail of the amplified face rectangle, you will see that it is not so easy to recognize a person.
So, at first glance, Us we recognize it from the context of the photo, but with only the rectangle, it is difficult to determine.
Well, the model has a similar problem, if you see a 50x50 face, it doesn't have enough information to determine a good descriptor.

This is the idea, but as I said we have not yet addressed the problem of resizing photos. You can still get the descriptor of a large face, on a much smaller rectangle, and theoretically its descriptor would not be good either.

However, I think we will agree that a 50px face is not good to compare. :sweat_smile:

lib/BackgroundJob/Tasks/CreateClustersTask.php
matiasdelellis: > But in general the rectangles are quite square, and if not, it is another poor quality alarm. thinking

Check my database, and the only faces that make a big difference are that they are either clipped on the edge, or they are just bad and have low confidence.

matiasdelellis: Done..

Issue #304, opened 2020-07-26
Removing .nomedia file does not trigger facerecognition when next analysis starts


lochbika:
### Expected behaviour

Removing the .nomedia file in a folder should enable facerecognition in next run

### Actual behaviour

Facerecognition does not run in folders after removal of .nomedia file

### Steps to reproduce
1. Create a new folder with some images and a .nomedia file in it
2. Run facerecognition -> folder is ignored
3. Remove the .nomedia file and run facerecognition again -> folder is still ignored during analysis

### Server configuration

* **Operating system:** Arch Linux

* **Pdlib version:**  latest release on Github

* **How is DLib installed:** From AUR, version 19.20
* **How is PDlib installed:** Manually 

* **PHP version:** 7.4.8

* **Web server:** nginx

* **Database:** mariadb

* **Nextcloud version:** 19.0.1

## Workaround
Running /usr/share/webapps/nextcloud/occ face:reset --image-error
and then repeat the facerecognition includes the previously ignored folders. However, this should be done automatically.

matiasdelellis:
Hi @lochbika 
Thank you very much for the report!.
This feature is implemented, but I have never tested it because we use our own folder setting file and just used this. :sweat_smile:  :disappointed: 

Please, test this change. https://github.com/matiasdelellis/facerecognition/pull/305

lochbika:
> Hi @lochbika
> Thank you very much for the report!.
> This feature is implemented, but I have never tested it because we use our own folder setting file and just used this. sweat_smile disappointed
> 
> Please, test this change. #305

Thanks for the quick reply! 
However, I would rather not try to install this git version, since I'm running my Nextcloud instance in a production setup, where the facerecognition is nearly finished (almost 15,000 photos analyzed, took 6 days). I rather not risk to compromise that.

Maybe someone else could test this.

Cheers

matiasdelellis:
> However, I would rather not try to install this git version, since I'm running my Nextcloud instance in a production setup,

Don't worry, when I saw the code it was pretty obvious that it was poorly implemented. :sweat_smile: 
So thanks again.

>  where the facerecognition is nearly finished (almost 15,000 photos analyzed, took 6 days)

I'm so embarrassed to tell you this after so much work, but hope you are using model 1 or 3, because lately we found that model 2 has problems and will be discouraged next week. :disappointed:

[...]

PR #305, opened 2020-07-26
Fix that removing .nomedia file does not trigger facerecognition when…

matiasdelellis:
… next analysis starts. Issue #304


lib/Watcher.php
```diff
@@ -220,7 +220,7 @@ public function postDelete(Node $node) {
 			// If user deleted file named .nomedia, that means all images in this and all child directories should be added.
 			// But, instead of doing that here, better option seem to be to just reset flag that image scan is not done.
 			// This will trigger another round of image crawling in AddMissingImagesTask for this user and those images will be added.
-			$this->settingsService->setNeedRemoveStaleImages(true, $owner);
+			$this->settingsService->setUserFullScanDone(false, $owner);
```
stalker314314, line 223: Good comment, wrong code:facepalm:

PR #309, opened 2020-07-30
Add face:migrate command and implement reset current model.

matiasdelellis:
It will be necessary to discard the model 2.. .


matiasdelellis:
Well, I admit that it is not the best code in the world, but I am going to merge it just like that. :wink:

PR #313, opened 2020-08-02
Introduce a new model v4: DLibCnnHogModel.

matiasdelellis:
Ok.  Basically it's a spawn. I recognize it. hahah 

Well, according to the last experiments, the weak link in our default model is
the landmark detector that can't align profile faces correctly to compare them
accurately.

So, In this model we detect the faces with CNN, but validate them with HOG.
That's the whole idea ..

The Hog detector also fails and cannot detect these faces.
So, if Hog detects it, we can assume that the landmark detector will do it too.
and this would be a good face..

If Hog don't detect this face, they are probably in profile or rotated.
These are bad to compare, so just we lower the confidence a little, to avoid
clustering them.

### DIFF
--- a/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
+++ b/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
@@ -226,7 +226,7 @@ class DlibCnnHogModel implements IModel {
 		return $this->fr->computeDescriptor($imagePath, $landmarks);
 	}
 
-	private function validateFace($proposedFace, $validataFaces) {
+	private function validateFace($proposedFace, $validateFaces) {
 		foreach ($validateFaces as $validateFace) {
 			$overlayPercent = $this->getOverlayPercent($proposedFace, $validateFace);
 			/**
@@ -278,7 +278,7 @@ class DlibCnnHogModel implements IModel {
 		$leftO = max($leftP, $leftV);
 		$rightO = min($rightP, $rightV);
 		$topO = max($topP, $topV);
-		$bottomO = min($bottomP, $botomV);
+		$bottomO = min($bottomP, $bottomV);
 
 		// Get area of both rect areas
 		$areaP = ($rightP - $leftP) * ($bottomP - $topP);
--- a/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
+++ b/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
@@ -255,16 +255,16 @@ class DlibCnnHogModel implements IModel {
 
 	private function getOverlayPercent($rectP, $rectV): float {
 		// Proposed face rect
-		$leftP = $rectP->getLeft();
-		$rightP = $rectP->getRight();
-		$topP = $rectP->getTop();
-		$bottomP = $rectP->getBottom();
+		$leftP = $rectP['left'];
+		$rightP = $rectP['right'];
+		$topP = $rectP['top'];
+		$bottomP = $rectP['bottom'];
 
 		// Validate face rect
-		$leftV = $rectV->getLeft();
-		$rightV = $rectV->getRight();
-		$topV = $rectV->getTop();
-		$bottomV = $rectV->getBottom();
+		$leftV = $rectV['left'];
+		$rightV = $rectV['right'];
+		$topV = $rectV['top'];
+		$bottomV = $rectV['bottom'];
 
 		// If one rectangle is on left side of other
 		if ($leftP > $rightV || $leftV > $rightP)
--- a/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
+++ b/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
@@ -146,6 +146,10 @@ class DlibCnnHogModel implements IModel {
 		$detectedFaces = [];
 
 		$cnnFaces = $this->dlibCnn5Model->detectFaces($imagePath);
+		if (count($cnnFaces) === 0) {
+			return $detectedFaces;
+		}
+
 		$hogFaces = $this->dlibHogModel->detectFaces($imagePath);
 
 		foreach ($cnnFaces as $proposedFace) {
@@ -171,11 +175,12 @@ class DlibCnnHogModel implements IModel {
 			 * can't align profile faces correctly.
 			 * The Hog detector also fails and cannot detect these faces.
 			 *
-			 * So, if Hog detects it (Overlay > 80%), we know that the landmark
+			 * So, if Hog detects it (Overlay > 35%), we can assume that landmark
 			 * detector will do it too.
-			 * Just return it.
+			 *
+			 * In this case, we consider the face valid, and just return it.
 			 */
-			if ($overlayPercent > 0.8) {
+			if ($overlayPercent >= 0.35) {
 				return $proposedFace;
 			}
 		}
--- a/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
+++ b/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
@@ -1,0 +1,292 @@
+<?php
+/**
+ * @copyright Copyright (c) 2020, Matias De lellis <mati86dl@gmail.com>
+ *
+ * @author Matias De lellis <mati86dl@gmail.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\FaceRecognition\Model\DlibCnnHogModel;
+
+use OCP\IDBConnection;
+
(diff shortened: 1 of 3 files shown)


matiasdelellis:
@stalker314314 
I haven't even tried to run it yet, but I'd like your opinion. :sweat_smile:

lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
```diff
@@ -0,0 +1,292 @@
...
+		$bottomP = $rectP['bottom'];
+
+		// Validate face rect
+		$leftV = $rectV['left'];
+		$rightV = $rectV['right'];
+		$topV = $rectV['top'];
+		$bottomV = $rectV['bottom'];
+
+		// If one rectangle is on left side of other
+		if ($leftP > $rightV || $leftV > $rightP)
```
stalker314314, line 270: in all of these checks, add `>=` instead of `>`. I think you can get division with zero on line 289 if `$rightP==$leftP` etc.

lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
```diff
@@ -0,0 +1,292 @@
...
+
+use OCA\FaceRecognition\Helper\MemoryLimits;
+
+use OCA\FaceRecognition\Service\FileService;
+use OCA\FaceRecognition\Service\ModelService;
+use OCA\FaceRecognition\Service\SettingsService;
+
+use OCA\FaceRecognition\Model\IModel;
+
+class DlibCnnHogModel implements IModel {
```
stalker314314, line 36: If I am not mistaken, this model is composition of two other models, so maybe it is better to have reference to them in this class (`private DlibHogModel $hogModel; private DlibCnnModel $cnnModel;`). You would not keep `private $cfd;` here (or others). Then, you would proxy all method to them. For example, `isInstalled` would return something like `$hogModel->isInstalled() && $cnnModel->isInstalled()`. `install`, `meetDependencies`, `getMaximumArea` etc similar logic. You would also forward call for `detectLandmarks` and `computeDescriptor` to `cnnModel`.

It was long time ago since I was learning design pattern:), but essentially I am describing https://en.wikipedia.org/wiki/Decorator_pattern

matiasdelellis: Ohh .. you're right .. I will try to do it later..

lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
matiasdelellis: Ohh .. you're right..

lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
matiasdelellis: Well .. It is not exactly what you expected, but I think for now it is enough ..

stalker314314: No, this is EXACTLY what I had in mind, every line of it:) Pure beauty now, IMHO:)

lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
```diff
@@ -0,0 +1,193 @@
...
+	public function getPreferredMimeType(): string {
+		return $this->dlibCnn5Model->getPreferredMimeType();
+	}
+
+	public function install() {
+		if ($this->isInstalled()) {
+			return;
+		}
+
+		// Insert on database and enable it
```
matiasdelellis, line 119: This is out of any objective of this PR, but it is something I have been thinking about.
The `facerecog_models` table is currently completely useless. I have my doubts for some future development, but now it could be removed without problems, since all the queries are made with Id configured in the SettingsService.

stalker314314: I guess...why not. I see its value as a way to have "foreign keys" and maybe in cases with complex upgrades/downgrades/custom models. But even any of this might not be an issue. Other than that, I am fine if you want to throw it away:)

matiasdelellis:
Well.. I think this may be a good model to recommend. but now is the moment of truth. :sweat_smile: 

In theory, we have to see all the faces of model 1, (Probably 50% more than in model 3), but without any bad cluster due to profile faces that usually appear in model.. 
In practice, I do not know... I'm running this model with the TBBT photos.. In principle it seems to take 7 hours. (Against 2.5hs of model 1, and 6hs of model 3). Much slower, but if it improves, it will be a good option.

Tomorrow when it finish, I will make a comparison of the four models.

matiasdelellis:
Ok..
I finish the analysis early, but as usual lately, it was a complete failure. :disappointed:  
Detect 1405 faces, and result on 1385 clusters. Basically nothing was clustered!! :sweat_smile: 

Wel, I didn't have to think long time to realize that 80% overlay is too little. Maybe I exaggerate a bit, but change that limit to 30%, and run it again.

Image Size | Images
-- | --
2048x1536 (4x3) | 433

As for the analysis time, it is 2 times slower that model 1.

433 Images | Model 1 | Model 2 | Model 3 | Model 4
-- | -- | -- | -- | --
AVG Time | 16,70 | 16,59 | 41,56 | 50,21
Procesing Time | 02:00:29 | 01:59:43 | 04:59:56 | 06:02:19
Percentage | 100,00 % | 99,37 % | 248,93 % | 300,71 %

As for the expected result, now it worked well! :grimacing: 
There was not even a single error in all the clusters. :smile: 

433 Images | Model 1 | Model 2 | Model 3 | Model 4
-- | -- | -- | -- | --
Faces | 1405 | 1405 | 1024 | 1405
Percentage | 100,00 % | 100,00 % | 72,88 % | 100,00 %
Faces in Bad Clusters | 13 | 63 | 0 | 0
Faces Error % | 0,93 % | 4,48 % | 0,00 % | 0,00 %

As for the number of clusters.. Increase the number of faces that wer not was clustered ( And it is exactly what was expected),

433 Images | Model 1 | Model 2 | Model 3 | Model 4
-- | -- | -- | -- | --
Total clusters | 757 | 730 | 423 | 819
Total clusters % | 100,00 % | 96,43 % | 55,88 % | 108,19 %
Clusters > 1 face | 83 | 102 | 54 | 66
Clusters > 1 % | 100,00 % | 122,89 % | 65,06 % | 79,52 %
Unclustered faces | 674 | 628 | 369 | 753
Unclustered % | 47,97 % | 44,70 % | 36,04 % | 53,59 %

Note that all percentages are against model 1. Maybe these don't have much sense. :sweat_smile: 

Well, In principle it seems to be working. Everyone will know if they want to use something so slow, but we should try it with a more real data set, to get an real idea if is an better model..

I guess this will be for the next release, since it will take me two weeks to do something similar with my photos, and I was hoping to release a release tomorrow..

p.s.: Also this reaffirms the fact that model 2 is bad. :wink:

matiasdelellis:
Well, last week I tried to do the same analysis between model 1 and model 4 with all personal photos. :grimacing: 

**Model 1**
Parameter | Absolut | Percent
-- | -- | --
Images (Model 1) | 10688 | 100,00 %
Total Faces | 9.820 | 100,00 %
Total clusters | 4617 | 100,00 %
  |   |  
Real Clusters | 531 | 11,50 %
Single Clusters | 4086 | 88,50 %
Bad Clusters | 45 | 0,97 %
  |   |  
Faces on Real Clusters | 5734 | 58,39 %
Faces on Bad Clusters | 352 | 3,58 %
  |   |  
Good clusters | 486 | 10,53 %
Good clustered faces | 5382 | 54,81 %

**Model 4**
Parameter | Absolut | Percent
-- | -- | --
Images (Model 4) | 10431 | 100,00 %
Total Faces | 9.584 | 100,00 %
Total clusters | 5320 | 100,00 %
  |   |  
Real Clusters | 380 | 7,14 %
Single Clusters | 4940 | 92,86 %
Bad Clusters | 3 | 0,06 %
  |   |  
Faces on Real Clusters | 4644 | 48,46 %
Faces on Bad Clusters | 12 | 0,13 %
  |   |  
Good clusters | 377 | 7,09 %
Good clustered faces | 4632 | 48,33 %

As you can see, model 4 was analyzed with 257 fewer photos. Basically, it was the last analysis I did, until I accidentally reset the entire database. (I canceled it sporadically, to reclustering and see how they evolve, and last time, exectute accidentally an old reset command.
 :disappointed: :disappointed_relieved:  **D'Oh!.**

(**NOTE:** We have to add a big confirmation warning to this command)

However, I still believe that it is very representative, since in percentage terms it remained stable among all analyzes. :wink: 

Ok. I summarize the parameters that IMHO are most important.

Parameter | Model 1 | Model 4
-- | -- | --
Good clusters | 10,53 % | 7,09 %
Good clustered faces | 54,81 % | 48,33 %
Bad real clusters | 8,47 % | 0,79 %
Faces on Bad Clusters | 3,58 % | 0,13 %

In short,
* **model 1:** 3.58% of errors.. and 10.53% of the clusters, handle the 54.81% of the faces, but 8.47% of real clusters have mixed faces.
* **model 4:** 0.13% of errors, and 7,09 % of the clusters, handle the 48.33% of the faces, but just 0.79% of real clusters have mixed faces.

So
* The errors are almost perfectly minimized, from 3.58% to 0.13%
* The number of real clusters is minimized. from 10.53% to 7.09. This probably doesn't say much.
* The amount of clustered faces is also minimized from 54% to 48%. This is 6.48% of faces that are not clustered. and may sound bad, but in IMHO it is compensated by minimizing errors.
* Bad real clusters (Calculated only against cluster with more thatone faces.) were also drastically minimized. This value really pleasantly surprised me.

matiasdelellis:
p.d's: 
Finally an experiment that more or less works well. :smile: :tada:  :fireworks: 

With that said, I'm going to merge it.  In principle, in the next release:
* Model 1 is still recommended.
* Model 2 is definitely deprecated.
* Model 3, it is recommended for low-resource devices.
* Model 4, This is a new model (2 times slower, but much more accurate.), which is in testing stages, and we invite you to test since probably will be the next recommended model.

PR #328, opened 2020-08-29
Add intitial funding...

matiasdelellis:
Sincerely, I'm not convinced to do this, but the short-term situation at my job justifies it. :thinking:


matiasdelellis:
If you want to join, add the links to the wiki. :wink:

stalker314314:
cool, go for it!

PR #336, opened 2020-09-08
Bye bye 'New person..'

matiasdelellis:
Ok..
This time I am not going to change the infrastructure beyond this (Just allow null names.). 

So, the main changes will be in the controllers, and the fronted. Then the persons will be all the clusters with the same name... And clusters, just that.

imagen

**TODO:**
- [x] Get the clusters without a name and suggest assigning one to them.
- [x] Rename all clusters on images of person view..
- [x] Change the sidebar panel considering in some way when it has a name and when it does not.
- [x] Fix and improve tests.

For the future we can continue with #262 and the comparisons would be between those who have names and those who do not.

### DIFF
--- a/appinfo/info.xml
+++ b/appinfo/info.xml
@@ -16,7 +16,7 @@
 - **🚀 Build your own thing:** FaceRecognition app is just a basic building block. Through FaceRecognition API, you can build your advanced scenarios - automatically add tags to images, connect contacts and persons, share images from specific person… We want to hear your ideas!
 ]]>
 	</description>
-	<version>0.6.3</version>
+	<version>0.6.4</version>
 	<licence>agpl</licence>
 	<author>Matias De lellis</author>
 	<author>Branko Kokanovic</author>
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -7,10 +7,16 @@ return ['routes' =>
 		'url'  => '/persons',
 		'verb' => 'GET'
 	],
-	// Get all clusters filtered by Name.
+	// Get all images filtered by Name.
 	[
 		'name' => 'person#findByName',
 		'url'  => '/person/{personName}',
+		'verb' => 'GET'
+	],
+	// Get all clusters filtered by Name.
+	[
+		'name' => 'person#findClustersByName',
+		'url'  => '/clusters/{personName}',
 		'verb' => 'GET'
 	],
 	// Get a cluster by Id.
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -27,6 +27,12 @@ return ['routes' =>
 	],
 	// Change name to a person.
 	[
+		'name' => 'person#updatePerson',
+		'url'  => '/person/{personName}',
+		'verb' => 'PUT'
+	],
+	// Change name to a cluster.
+	[
 		'name' => 'person#updateName',
 		'url'  => '/cluster/{id}',
 		'verb' => 'PUT'
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -4,7 +4,7 @@ return ['routes' =>
 	// Get all face clusters with faces and file asociated.
 	[
 		'name' => 'person#index',
-		'url'  => '/clusters',
+		'url'  => '/persons',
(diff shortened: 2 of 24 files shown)


matiasdelellis:
p.s. The idea is basically to finish some discussions about join clustering and performance. :sweat_smile:

escoand:
Good idea. And this is also the kind of view you want to integrate into images app?

matiasdelellis:
Hi @escoand 

> Good idea. And this is also the kind of view you want to integrate into images app?

Yes. Hope to have something like this. However, there is no api for it yet. :wink:

matiasdelellis:
imagen

:wink:

matiasdelellis:
Well, for the main view this would look more or less like this..

**Main view:** Persons
imagen

**When selecting a person:** Show all photos of all clusters with the same name. Note that reuses the nextcloud thumbnails and therefore the server load is less.
imagen

If the "person" has multiple clusters, it shows the button to see them and allows rename each clusters.

**Clusters view**
imagen

**TODO:**
- [ ] Get the clusters without a name and suggest assigning one to them.
- [ ] Change the sidebar panel considering in some way when it has a name and when it does not.
- [ ] Fix and improve tests.

For the future we can continue with #262 and the comparisons would be between those who have names and those who do not.

escoand:
OK, in the first layer you have added the "person abstraction" and show only ambigous names, right? Thats effectively the often requested merge of multiple clusters with really less overhead and does not change the background work. Simple but nice solution.

But I don't understand why you show only one cluster on the second layer. Wouldn't make it more sense to show already here all clusters but allow to rename them?

But I would think the separation and rename button should be only small as this is in most cases just an onetime action.

matiasdelellis:
HI @escoand 

> But I don't understand why you show only one cluster on the second layer.

They are also all clusters of one person. Count the pictures in the second and third pictures. :wink: 

> Couldn't make it more sense to show already here all clusters but allow to rename them?

You are right, also still need to rename all the clusters of the same person in the second view. Soon... :grimacing:

The third view is still necessary to fix grouping errors since the most common mistake is sometimes not using the last names, and 'join' two persons with the same name.. :sweat_smile:

matiasdelellis:
When there are already unnamed clusters ... (I'm reusing the old dialog. Maybe make another ...)

imagen

matiasdelellis:
With last commit change url when you select a person.
e.g. `https://nextcloud.com/settings/user/facerecognition?name=Sheldon%20Cooper`

So, You can put as favorites in your browser and access the photos of your favorite people with it.. :grimacing: 

@dassio, this is especially for #344

dassio:
>  e.g. `https://nextcloud.com/settings/user/facerecognition?name=Sheldon%20Cooper`

shouldn't it be `https://nextcloud.com/index.php/settings/user/facerecognition?name=Sheldon%20Cooper` ?

matiasdelellis:
> > e.g. https://nextcloud.com/settings/user/facerecognition?name=Sheldon%20Cooper
>
> shouldn't it be https://nextcloud.com/index.php/settings/user/facerecognition?name=Sheldon%20Cooper ?

Yes... The public url depends on the RewriteBase configuration, but the api itself must include the `index.php`, however don't worry about it. Just trust on `OCP\IURLGenerator`.

matiasdelellis:
@dassio in last commit add an method to get this.

I will merge this patch tomorrow. Wait and reuse all this code. :smile:

DerOetzi:
I'm very excited about this feature. When will this be available for update. Or is there the possibility to manual update to version 0.6.4?

matiasdelellis:
Hi @DerOetzi 
Thanks, We hope it meets expectations.. :sweat_smile: 

Surely in a couple of days I'll release the update. I just hope to merge #344 and need testers on NC20 for that. :wink:

DerOetzi:
@matiasdelellis if you can give me a hint how to test I can give it a try on my instance.

matiasdelellis:
Hi @DerOetzi 
See https://github.com/matiasdelellis/facerecognition/issues/343#issuecomment-713011868

Issue #341, opened 2020-10-02
Crashed on postWrite and prevent other apps to work


GAS85:
### Expected behaviour

Face Recognition App should not crash on post write

### Actual behaviour

Its crashed and gives me no opportunity to use auto export of Phone Track.
Issue for reference: https://gitlab.com/eneiluj/phonetrack-oc/-/issues/361

### Steps to reproduce
1. Install Face Recognition and Phone Track
2. Setup Auto export of Phone Track
3. Check that it does not works and you have following in logs:
```
[...]

PR #342, opened 2020-10-03
Ignore file events when was not emitted by users.. Issue #341

matiasdelellis:


stalker314314:
For my own education - how can this happen?:) If file is changed/added, does it matter if user is the one doing it or not? I mean - file is changed, we should react to it, right?

matiasdelellis:
Hi @stalker314314 

> I mean - file is changed, we should react to it, right

Yes!. I full agree..
To be honest, just look for the easiest solution, however, I think the event should be emited with the proper context but it is something that we cannot control.

So:
 * We can accept this change. (Note that the original problem is similar to thumbnails), but certainly don't know if it causes more problems. 😕
 * Or I can fix just this report just by checking the mimetype before the user, but there may be some other application that does the same with jpeg, and we would return here .. 🤔

Although I do not solve the underlying problem, maybe the second option is better .. No?

stalker314314:
So., problem is, IIUC, that we cannot deduce `$owner`, right? But `$node` should point to owner, it is usually in filepath? Maybe we should fallback to some other way to deduce who is owner of the file being changed? I am not sure... If not (or if it is ugly hack), maybe it is not worth it and your solution is OK.

matiasdelellis:
> So., problem is, IIUC, that we cannot deduce $owner, right?

Yes. sorry if not clarify it.. 😅 

> But $node should point to owner

Yes, but us we use the logged user basically to appropriate the shared files, since the node reports the original owner.

Maybe we can revert it with some improvements: https://github.com/matiasdelellis/facerecognition/commit/a42f2e46abf32fbbfa25f5e2e4bd9819cbb2c62b#diff-6ff6cba3b61bf5ad548b5cc1f02b094a

> it is usually in filepath?

If you mean to go back to this, https://github.com/matiasdelellis/facerecognition/commit/927e836a5e1036e353ee2f54e2819d3234ebd432#diff-6ff6cba3b61bf5ad548b5cc1f02b094a
.. I always found a very fragile code and i don't know how it works with shared files. 

> Maybe we should fallback to some other way to deduce who is owner of the file being changed?

I liked doing a mixed solution by partially reversing the first commit.

stalker314314:
Previous solution (`$node->getOwner()`) looks like a proper way to get owner of a file, at least for me as a amateur in this code.

> I always found a very fragile code and i don't know how it works with shared files.

You removed because you found some bug with it, or just for "FUD" (fear, uncertainty, doubt)? I know you well enough not to think you would remove code just like that:) There was probably something wrong with this code?
I understand this approach may not work with shared files (if file is changed, all users with access to this file need to be be notified, not just owner or logged user), but I think neither of those 2 solutions address this problem (which is deeper than this change).

> I liked doing a mixed solution by partially reversing the first commit.

Yes, seems like mixed solution (fallback to old behavior) could do the trick!

matiasdelellis:
Close in favour of #346

PR #345, opened 2020-10-11
bump nextcloud version to 20 

dassio:
CI BUILD: https://travis-ci.com/github/dassio/facerecognition/builds/189297176

Edit:
i also did an manual test, it seems to be ok

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,10 @@ env:
     - DB=mysql CORE_BRANCH=stable19
     - DB=pgsql CORE_BRANCH=stable19
     - DB=sqlite CORE_BRANCH=stable19
+    - DB=mysql CORE_BRANCH=stable20
+    - DB=pgsql CORE_BRANCH=stable20
+    - DB=sqlite CORE_BRANCH=stable20
+
 
 before_install:
   # Installing pdlib (install dlib from repo and compile pdlib on top of it)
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -20,9 +20,9 @@ env:
     - DB=mysql
     - DB=pgsql
     - DB=sqlite
-    - DB=mysql CORE_BRANCH=stable19
-    - DB=pgsql CORE_BRANCH=stable19
-    - DB=sqlite CORE_BRANCH=stable19
```
matiasdelellis, line 25: Don't delete NC19.. Just add NC20.

Currently the stable versions are NC18 to NC20, so we must test all 3. :wink:

matiasdelellis (changes requested):
Hi @dassio 
Thank you very much for split the patch..   Returns the test of NC 19 and we merge it. :smile:

dassio:
> p.s: I was worried about the new Bootstraping process but it seems optional. https://docs.nextcloud.com/server/latest/developer_manual/app_development/bootstrap.html#bootstrapping

yes, i am trying to find a way  so that we can active the unified search but also be compatible with NC19

matiasdelellis:
> yes, i am trying to find a way so that we can active the unified search but also be compatible with NC19

Yes,, Please.. :smile: 

See:
* https://github.com/nextcloud/deck/pull/2200
* https://github.com/nextcloud/deck/blob/master/lib/AppInfo/Application.php

matiasdelellis:
Thanks again.

PR #367, opened 2020-10-31
Fix inconsistent wording in settings page

Derkades:
a few things I noticed while translating

### DIFF
--- a/templates/settings/admin.php
+++ b/templates/settings/admin.php
@@ -56,7 +56,7 @@
 		</p>
 		<br>
 		<h3>
-			<?php p($l->t('Model'));?>
+			<?php p($l->t('Configuration information'));?>
 			<span class="status success<?php if(!($_['model-version'] > 0 && $_['meet-dependencies'])):?> error<?php endif;?>"></span>
 		</h3>
 		<p><?php p($l->t('Current model:'));?> <em><?php p($_['model-version']);?></em></p>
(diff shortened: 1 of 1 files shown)


matiasdelellis:
Hi @Derkades 
Thank you very much for these corrections.
In general, I think you are right, but if you are not offended, I would rather postpone this PR for later. I want to release a minor revision these days, and if I apply this patch, there will be many incomplete translations. Then we can give the rest of the translators more time to update.

templates/settings/admin.php
```diff
@@ -47,19 +47,19 @@
...
 			<span><input type="range" id="min-confidence-range" min="0.0" max="1.1" value="0.99" step="0.01" class="ui-slider"></span>
-			<label for="min-confidence-range"><?php p($l->t('Highest minimum confidence'));?></label>
+			<label for="min-confidence-range"><?php p($l->t('Higher minimum confidence'));?></label>
 			<span id="min-confidence-value"class="span-highlighted">...</span>
 			<a id="restore-min-confidence" class="icon-align icon-history" style="display: none;" title="<?php p($l->t('Restore'));?>" href="#"></a>
 			<a id="save-min-confidence" class="icon-align icon-edit" style="display: none;" title="<?php p($l->t('Save'));?>" href="#"></a>
 		</p>
 		<br>
 		<h3>
-			<?php p($l->t('Configuration information'));?>
```
matiasdelellis, line 59: Here should be more information, such as the version of pdlib, if CUDA it is enabled, etc. So, don't change it. :wink:

Derkades:
Of course, merge this PR at any time you want or not at all.

matiasdelellis:
Thanks again @Derkades

Derkades:
These translations haven't updated for me on transifex, am I missing something?

matiasdelellis:
Hi @Derkades 
Tomorrow add that.. :wink:

Issue #364, opened 2020-10-28
How to scan Groupfolders?


trvrcr:
### Expected behaviour

When starting a scan, not only the personal folders but also the groupfolders.
I tried it by enabling the following options:
config:app:set facerecognition handle_shared_files --value true

### Actual behaviour

nothing happens

### Server configuration

* **Operating system:** Raspbian (on Raspberry Pi4  4GB got it working)

* **PHP version:** 7.4

* **Web server:** Nginx 1.18

* **Database:**Mariadb 10.3

* **Nextcloud version:** 20.0.1

### Client configuration

* **Browser:** Mozilla 82.0

* **Operating system:** Windows

#### Background task log with debug.
<details>
<summary>sudo -u apache php occ -vvv face:background_job</summary>

```
1/10 - Executing task CheckRequirementsTask (Check all requirements)
        System: Linux
        System memory: 3562459136
        PHP Memory Limit: 2146435072
2/10 - Executing task CheckCronTask (Check that service is started from either cron or from command)
3/10 - Executing task LockTask (Acquire lock so that only one background task can run)
4/10 - Executing task DisabledUserRemovalTask (Purge all the information of a user when disable the analysis.)
yielding
yielding
yielding
yielding
yielding
5/10 - Executing task StaleImagesRemovalTask (Crawl for stale images (either missing in filesystem or under .nomedia) and remove them from DB)
        Skipping stale images removal for user XXX as there is no need for it
        Skipping stale images removal for user XXX as there is no need for it
        Skipping stale images removal for user XXX as there is no need for it
        Skipping stale images removal for user XXX as there is no need for it
        Skipping stale images removal for user XXX as there is no need for it
6/10 - Executing task CreateClustersTask (Create new persons or update existing persons)
        Skipping cluster creation, not enough data (yet) collected. For cluster creation, you need either one of the following:
        * have 1000 faces already processed
        * or you need to have 95% of you images processed
        Use stats command to track progress
yielding
        Skipping cluster creation, not enough data (yet) collected. For cluster creation, you need either one of the following:
        * have 1000 faces already processed
        * or you need to have 95% of you images processed
        Use stats command to track progress
yielding
        Skipping cluster creation, not enough data (yet) collected. For cluster creation, you need either one of the following:
        * have 1000 faces already processed
        * or you need to have 95% of you images processed
        Use stats command to track progress
yielding
        Found 0 faces without associated persons for user XXX and model 2
[...]

PR #368, opened 2020-10-31
Add support to handle Group folders

matiasdelellis:
Fix issue #364


matiasdelellis:
[facerecognition.tar.gz](https://github.com/matiasdelellis/facerecognition/files/5469163/facerecognition.tar.gz)

Issue #371, opened 2020-11-02
Change order of face recognition


cliffalbert:
This is more of a feature request then a bug. But it would be nice to first check for new image/faces and after that do the createclusterstask. 

I think this behaviour was build this way due to the face recognition running from a cron job. But in my case it is usually run manually.

The problem is that when first running CreateClusterTask and then looking for new images is that in practice you have to run the background job twice before pictures start to show up in the face recognition web interface.

Maybe this behaviour could be introduced using an addition flag ? Like --check-new-first ?

matiasdelellis:
Hi @cliffalbert 

It is in consideration.
In general, I would think about switching these tasks when time is not indicated. So, if you indicate time, it is because you have it as a scheduled task, and If you did not indicate a time limit, it is because at the end you expect to see the result when finish the command.

However, you will understand that it is not within the priorities. 🙈 
Thanks for the report. 😬

cliffalbert:
Let me fiddle around with changing the order when timeout is 0, as far as I can see it wouldn't be that hard to implement. If I get it working correctly expect a PR ;)

matiasdelellis:
Great @cliffalbert 
I want to wait for a comment from Branko but it looks good.. Thanks.

matiasdelellis:
Hi @cliffalbert
Merged on master... 
Thanks for all.. :wink:

PR #374, opened 2020-11-03
Run tasks in optimized order when timeout = 0. This suggests starting from command line

cliffalbert:
This is a feature implementation for issue #371

### DIFF
--- a/lib/BackgroundJob/BackgroundService.php
+++ b/lib/BackgroundJob/BackgroundService.php
@@ -89,18 +89,33 @@ class BackgroundService {
 
 		// Here we are defining all the tasks that will get executed.
 		//
-		$task_classes = [
-			CheckRequirementsTask::class,
-			CheckCronTask::class,
-			LockTask::class,
-			DisabledUserRemovalTask::class,
-			StaleImagesRemovalTask::class,
-			CreateClustersTask::class,
-			AddMissingImagesTask::class,
-			EnumerateImagesMissingFacesTask::class,
-			ImageProcessingTask::class,
-			UnlockTask::class
-		];
+		if ($timeout > 0) {
+			$task_classes = [
+				CheckRequirementsTask::class,
+				CheckCronTask::class,
+				LockTask::class,
+				DisabledUserRemovalTask::class,
+				StaleImagesRemovalTask::class,
+				CreateClustersTask::class,
+				AddMissingImagesTask::class,
+				EnumerateImagesMissingFacesTask::class,
+				ImageProcessingTask::class,
+				UnlockTask::class
+			];
+		} else {
+                       $task_classes = [
+				CheckRequirementsTask::class,
+				CheckCronTask::class,
+				LockTask::class,
+				DisabledUserRemovalTask::class,
+				StaleImagesRemovalTask::class,
+				AddMissingImagesTask::class,
+				EnumerateImagesMissingFacesTask::class,
+				ImageProcessingTask::class,
+				CreateClustersTask::class,
+				UnlockTask::class
+                        ];
+		};
 
 		// Main logic to iterate over all tasks and executes them.
 		//


matiasdelellis:
@stalker314314 Since it's all your magic .. Do you have any comments?

lib/BackgroundJob/BackgroundService.php
```diff
@@ -89,18 +89,33 @@ public function execute(int $timeout, bool $verbose, IUser $user = null, int $ma
...
-			CheckCronTask::class,
-			LockTask::class,
-			DisabledUserRemovalTask::class,
-			StaleImagesRemovalTask::class,
-			CreateClustersTask::class,
-			AddMissingImagesTask::class,
-			EnumerateImagesMissingFacesTask::class,
-			ImageProcessingTask::class,
-			UnlockTask::class
-		];
```
matiasdelellis, line 103: Introduce some var to make it more understandable ..

Maybe:

```
$singleTaskOrder = ($timeout > 0);
if ($singleTaskOrder) {
```

But the name doesn't convince me either .. Any other suggestions? :thinking:

stalker314314: yes, maybe some comment why order is different. And you should probably document it.

lib/BackgroundJob/BackgroundService.php
```diff
@@ -89,18 +89,33 @@ public function execute(int $timeout, bool $verbose, IUser $user = null, int $ma
...
+		if ($timeout > 0) {
```
stalker314314, line 92: Not sure if this should be coupled with `timeout` or there should be another argument, something like `--favor-create` (this is stupid, but not sure what is better name here anyway)...

Derkades:
I do not need this new order but in my opinion it can be done this way in all cases, not just when timeout == 0. It would make the code simpler and I can't think of a situation where cron job users want to keep the current task order.

matiasdelellis:
@Derkades 

No. The reason for the original order is that there are tasks of "infinite" execution time  (Of course not infinite, but it takes as long as the number of photos you have!).
So if you use the task as scheduled with the order reversed, ImageProcessingTask easily takes longer than the timeout, and therefore the cluster creation will never run!.

lib/BackgroundJob/BackgroundService.php
matiasdelellis: Maybe `--clustering-last` :thinking:

strugee: `--defer-clustering` maybe?  I like that a little better since it seems more like a command ("my intentions are to _defer clustering_") but maybe that's just me. And not that it really matters anyway 😅

matiasdelellis: Great name.. Do you want to do it?.. 😬 
Now I add some comments that could help you. 😉

matiasdelellis:
https://github.com/matiasdelellis/facerecognition/blob/master/lib/Command/BackgroundCommand.php#L76
```
			)->addOption(
				'defer-clustering',
				null,
				InputOption::VALUE_NONE,
				'change the order of tasks to get the cluster at the end', // Adjust the comment. sure you write it better.
				null
			);
```

Bellow... 
```
$deferClustering = false;
if ($input->getOption('defer-clustering')) {
            $deferClustering = true;
}
```

And finaly...
```
--- $this->backgroundService->execute($timeout, $verbose, $user, $maxImageArea);
+++ $this->backgroundService->execute($timeout, $verbose, $user, $maxImageArea, $deferClustering);
```

I guess the next is more obvious ..., 😬

matiasdelellis:
p.s: ...and document it in the readme .. 😉

cliffalbert:
Will update this this week, thanks for the feedback. Been very busy with work, so no time to look at this ;)

matiasdelellis:
Hi @cliffalbert 
I appropriate this PR and continue it in #410 
Please see that.. :grimacing:

PR #390, opened 2020-11-25
✨ Allow Setting a Custom Model Path ✨

hahagu:
Adds an option to set a custom model path.

### DIFF
--- a/lib/Service/ModelService.php
+++ b/lib/Service/ModelService.php
@@ -27,6 +27,7 @@ use OCP\IConfig;
 use OCP\Files\IAppData;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
+use OCA\FaceRecognition\Service\SettingsService;
 
 class ModelService {
 
@@ -48,16 +49,33 @@ class ModelService {
 		$this->config     = $config;
 		$this->appData    = $appData;
 		$this->rootFolder = $rootFolder;
-
-		// Construct root folder for models
-		$this->prepareAppDataFolders();
+		
+		// Set model path.
+		setModelPath();
 
 		/// Get this folder.
 		$instanceId = $this->config->getSystemValue('instanceid', null);
 		$appData = $this->rootFolder->get('appdata_'.$instanceId)->getPath();
 		$dataDir = $this->config->getSystemValue('datadirectory', null);
+	}
 
-		$this->modelsFolder = $dataDir . $appData . '/facerecognition/models/';
+	/**
+	 * @return void
+	 */
+	private function setModelPath() {
+		$modelPath = $this->settingsService->getModelPath();
+
+		// Check if the modelPath value is set
+		if (is_null($modelPath)) {
+			// Set model folder to set path
+			$this->modelsFolder = $modelPath;
+		} else {
+			// Construct root folder for models
+			$this->prepareAppDataFolders();
+
+			// Set default model folder
+			$this->modelsFolder = $dataDir . $appData . '/facerecognition/models/';
+		}
 	}
 
 	/**
@@ -79,9 +97,9 @@ class ModelService {
 	 */
 	public function prepareModelFolder(int $modelId) {
 		try {
-			$this->appData->getFolder('/models/' . $modelId);
+			$this->rootFolder->getFolder($this->modelsFolder . $modelId);
 		} catch (NotFoundException $e) {
-			$this->appData->newFolder('/models/' . $modelId);
+			$this->rootFolder->newFolder($this->modelsFolder . $modelId);
 		}
 	}
 
--- a/lib/Service/ModelService.php
+++ b/lib/Service/ModelService.php
@@ -50,6 +50,9 @@ class ModelService {
 		$this->appData    = $appData;
 		$this->rootFolder = $rootFolder;
 		$this->modelsFolder = $this->settingsService->getModelPath();
+
+		// Prepare Model Folder
+		prepareModelFolder();
 	}
 
 	/**
@@ -71,19 +74,7 @@ class ModelService {
 	 */
 	public function prepareModelFolder(int $modelId) {
 		if (!is_dir($this->modelsFolder)) {
-			mkdir($this->modelsFolder);
+			mkdir($this->modelsFolder, 0770, true);
 		}
 	}
-
-	/**
-	 * @return void
-	 */
-	private function prepareAppDataFolders() {
-		try {
-			$this->appData->getFolder('/models');
-		} catch (NotFoundException $e) {
-			$this->appData->newFolder('/models');
-		}
-	}
-
 }
--- a/lib/Service/ModelService.php
+++ b/lib/Service/ModelService.php
@@ -49,10 +49,30 @@ class ModelService {
 		$this->config     = $config;
 		$this->appData    = $appData;
 		$this->rootFolder = $rootFolder;
-		$this->modelsFolder = $this->settingsService->getModelPath();
+		$this->modelsFolder = getModelPath();
 
 		// Prepare Model Folder
 		prepareModelFolder();
+	}
+
+	/**
+	 * @return string
+	 */
+	private function getModelPath(): string {
+		// Get model path setting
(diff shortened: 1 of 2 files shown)


lib/Service/SettingsService.php
```diff
@@ -57,6 +57,9 @@ class SettingsService {
 	const DEFAULT_SENSITIVITY = '0.4';
 	const MAXIMUM_SENSITIVITY = '0.6';
 
+	/** Model Path used for read and writing models  **/
+	const SYSTEM_MODEL_PATH = 'modelPath';
```
matiasdelellis, line 61: const SYSTEM_MODEL_PATH = 'modelPath';
```suggestion
	const SYSTEM_FACE_RECOGNITION_MODEL_PATH = 'FaceRecognitionModelPath';
```

matiasdelellis: Keep in mind that this configuration will go in `config/config.php` and it should be clear.

lib/Service/SettingsService.php
```diff
@@ -240,6 +243,10 @@ public function setMinimumConfidence($confidence) {
 	 * The next settings are advanced preferences that are not available in gui.
 	 * See: https://github.com/matiasdelellis/facerecognition/wiki/Settings#hidden-settings
 	 */
+	public function getModelPath(): string {
+		return $this->config->getSystemValue(self::SYSTEM_MODE_PATH, null);
+	}
+
```
matiasdelellis, line 249: Move it here: https://github.com/matiasdelellis/facerecognition/pull/390/files#diff-0d0e5b28c99581ae724962d51121ee75bb2552319070923c2aa9112dab4f2e74R303

..with the appropriate comment. :wink:

lib/Service/ModelService.php
```diff
@@ -79,9 +97,9 @@ public function modelFileExists(int $modelId, string $file): bool {
 	 */
 	public function prepareModelFolder(int $modelId) {
 		try {
-			$this->appData->getFolder('/models/' . $modelId);
```
matiasdelellis, line 82: Don't use root folder.. It can be literally any directory that has the correct permissions.

So, revert and add an:
```
if (!is_null($modelPath)) {
    mkdir($modelPath)
    return;
}
```

lib/Service/ModelService.php
```diff
@@ -48,16 +49,33 @@ public function __construct(IConfig     $config,
...
 		$instanceId = $this->config->getSystemValue('instanceid', null);
 		$appData = $this->rootFolder->get('appdata_'.$instanceId)->getPath();
 		$dataDir = $this->config->getSystemValue('datadirectory', null);
+	}
 
-		$this->modelsFolder = $dataDir . $appData . '/facerecognition/models/';
+	/**
+	 * @return void
+	 */
+	private function setModelPath() {
```
matiasdelellis, line 65: Rename `getModelPath()` that return string, and use on constructor.

```
private function getModelPath(): string {
    if (!is_null($modelPath)) {
       return $modelPath;
    }

    $instanceId = $this->config->getSystemValue('instanceid', null);
    $appData = $this->rootFolder->get('appdata_'.$instanceId)->getPath();
    $dataDir = $this->config->getSystemValue('datadirectory', null);

    return $dataDir . $appData . '/facerecognition/models/';
}
```

matiasdelellis (changes requested):

lib/Service/ModelService.php
matiasdelellis:
```
if (!is_null($modelPath)) {
    if (file_exists($modelPath)) {
        mkdir($modelPath)
    }
    return;
}
```

MMM.. 
On second thought, to avoid errors ..

lib/Service/ModelService.php
hahagu: What is the prepareAppdataFolders function for?
it seems to create a 'model' directory on the appdata_{id}/models

hahagu:
This should fix the issues you mentioned! 😄

lib/Service/ModelService.php
matiasdelellis: > What is the prepareAppdataFolders function for?
> It seems to create a 'model' directory under appdata_{id}.
> However, shouldn't it be under appdata_{id}/facerecognition?

Yes.. it's just to create the folder 'facerecognition/models' within appData.

matiasdelellis:
```
private function getModelPath(): string {
    if (!is_null($modelPath)) {
       return $modelPath;
    }

    try {
        $this->appData->getFolder('/models');
    } catch (NotFoundException $e) {
         $this->appData->newFolder('/models');
    }

    $instanceId = $this->config->getSystemValue('instanceid', null);
    $appData = $this->rootFolder->get('appdata_'.$instanceId)->getPath();
    $dataDir = $this->config->getSystemValue('datadirectory', null);

    return $dataDir . $appData . '/facerecognition/models/';
}
```

hahagu: > Yes.. it's just to create the folder 'facerecognition/models' within appData.

I believe that needs to be addressed then?

lib/Service/SettingsService.php
```diff
@@ -240,6 +243,23 @@ public function setMinimumConfidence($confidence) {
 	 * The next settings are advanced preferences that are not available in gui.
 	 * See: https://github.com/matiasdelellis/facerecognition/wiki/Settings#hidden-settings
 	 */
+	private function getModelPath(): string {
+		// Get model path setting
```
matiasdelellis, line 247: NO!.
here just return the value as before.. The differentiation between both folders must be done in the Model Service!

lib/Service/ModelService.php
hahagu: Done!
Recursive and permissions 0770 added.
The default 0777 seems unnecessarily open.

lib/Service/ModelService.php
```diff
@@ -78,22 +73,8 @@ public function modelFileExists(int $modelId, string $file): bool {
 	 * @return void
 	 */
 	public function prepareModelFolder(int $modelId) {
-		try {
-			$this->appData->getFolder('/models/' . $modelId);
-		} catch (NotFoundException $e) {
-			$this->appData->newFolder('/models/' . $modelId);
```
matiasdelellis, line 84: I meant to revert your change, and just add that .. The creation of the models /MODEL_ID folder should be done here.

lib/Service/SettingsService.php
matiasdelellis: I meant to just change the getModelfolder function here. Not all new code.. This doesn't even work, because you don't have rootFolder, etc.

matiasdelellis (changes requested):

lib/Service/ModelService.php
hahagu: What if I do
`public function prepareModelFolder(int $modelId) {
		if (!is_dir($this->modelsFolder)) {
			mkdir($this->modelsFolder . '/' . $modelID, 0770, true);
		}
	}
`

matiasdelellis: > mkdir($this->modelsFolder . '/' . $modelID, 0770, true);

You are right you should use this in case of custom folder.. :sweat_smile: 

But in the default case, you should create the folder using the AppData methods,  because nextcloud sve this in the database and may fail.

matiasdelellis (changes requested):

lib/Service/SettingsService.php
matiasdelellis: D'OH!, Again.

hahagu:
If custom model path is set,
try and catch of
`$this->appData->newFolder('/facerecognition/models/' . $modelId);`
is run instead of the external directive mkdir. 😀
Hope this solves everything?

matiasdelellis:
Hi @hahagu 
Sorry, but it still didn't even work, and it was taking longer than necessary.. :disappointed: 

Close in favour to #391

mgutschmidt:
Hi, can you describe how to use this Setting? I also use S3 Storage, but can't figure out to get this solution running...

PR #391, opened 2020-11-25
Allow Setting a Custom Model Path. Replaces #390

matiasdelellis:
Thanks you so much @hahagu

### DIFF
--- a/lib/Service/ModelService.php
+++ b/lib/Service/ModelService.php
@@ -28,6 +28,8 @@ use OCP\Files\IAppData;
 use OCP\Files\IRootFolder;
 use OCP\Files\NotFoundException;
 
+use OCA\FaceRecognition\Service\SettingsService;
+
 class ModelService {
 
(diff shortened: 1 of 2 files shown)


hahagu:
> Thanks you so much @hahagu

I think it caused you more trouble than doing it yourself 😅
Thanks for helping!
I've really only provided the idea for it :)

mgutschmidt:
Hi, can you describe how to use this Setting? I also use S3 Storage, but can't figure out to get this solution running...

hahagu:
> Hi, can you describe how to use this Setting? I also use S3 Storage, but can't figure out to get this solution running...

https://github.com/matiasdelellis/facerecognition/blob/9b4ebc9e8eab169444653c90d87237a8ce2b4561/lib/Service/SettingsService.php#L133

mgutschmidt:
Ahh, too simple^^. It worked, thanks!

Issue #429, opened 2021-02-14
Nextcloud 21 compatibility


Derkades:
Issue for tracking Nextcloud 21 compatibility. I have not tried upgrading and force enabling it yet.

dblazquezl:
Hi, I already installed it with version 21, still working in the initial process of all my photos. It will take some time.

I just need to modify a couple of thing in the dockerfile in order to get the image:

1.- bz2 was not present, so I copied the 2 lines installing it to the end of the dockerfile instead in the intermediate image and then 
     it worked.
2.- change this path to the latest version:
      COPY --from=builder /usr/local/lib/php/extensions/no-debug-non-zts-20190902/pdlib.so /usr/local/lib/php/extensions/no- 
      debug-non-zts-20190902/
3.- add backslash at the end of this line (it got an error without it)
      COPY --from=builder /usr/local/lib/libdlib.so* /usr/local/lib/


I think that was all, I will try to update when the initial process of the images finish

markuman:
I've  moved from NC 20 php 7.4 to NC21 php 8.0 and it seems there is an issue with the temporary path.


```
Faces found: 0. Image will be skipped because of the following error: Error during image resize
```

any ideas? @matiasdelellis

markuman:
I've patched the upstreamd nextcloud docker image like that: https://github.com/nextcloud/docker/issues/1421#issuecomment-783499435  
It's build as `nextcloud21`.  
And this is my multi-stage build on that for facerecognition dependencies: https://git.osuv.de/m/osuv/src/branch/main/roles/containers/files/nextcloud.Dockerfile

You can pull it from here `registry.gitlab.com/markuman/container-images:home.21.0.0.1`

markuman:
Rolling back to php 7.4 works.

jmigual:
Hi @markuman I just used your Docker file and the PDLIB tests failed.

I got the following error:

<details>
<summary>Log</summary>

```
----------------------------------------------------------------------
Libraries have been installed in:
   /var/www/html/pdlib/modules

If you ever happen to want to link against installed libraries
in a given directory, LIBDIR, you must either use libtool, and
specify the full pathname of the library, or use the `-LLIBDIR'
flag during linking and do at least one of the following:
   - add LIBDIR to the `LD_LIBRARY_PATH' environment variable
     during execution
   - add LIBDIR to the `LD_RUN_PATH' environment variable
     during linking
[...]

PR #437, opened 2021-03-13
allow max-nc-version 21 and max-php-version 7.4

matiasdelellis:
Since it was by several users confirmed in #429 that the app works on NC21 with PHP 7.4, I've just took the opportunity to add those values.
BTW: afaik, since a while the appstore should actually respect max versions for php before updating/installing the app.

@szaimen It seems to work except for the php 7.2 exclusion in NC21 :grimacing: :thinking: 

Thanks.. :smile:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,21 +15,19 @@ php:
   
 env:
   global:
-    - CORE_BRANCH=stable18
+    - CORE_BRANCH=stable20
     - APP_NAME=facerecognition
   matrix:
     - DB=mysql
     - DB=pgsql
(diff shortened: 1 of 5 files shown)


szaimen:
4a8b0d6 looks good 👍 based on:
Screenshot_20210313-144659_Brave

matiasdelellis:
https://travis-ci.org/github/matiasdelellis/facerecognition/builds/762723293

Great. It worked. :smiley: 
Now I have to investigate the last reported in #429 :grimacing: 

Thanks again @szaimen

PR #441, opened 2021-03-17
Change the Sensitivity setting to Clustering Threshold.

matiasdelellis:
The term sensitivity was intended to be less technical about how
it works, but it turned out to be more confusing for users.

This is only on the fronted, and the backend will not change this
time.

### DIFF
--- a/templates/settings/admin.php
+++ b/templates/settings/admin.php
@@ -26,15 +26,16 @@
 		</p>
 		<br>
 		<h3>
-			<?php p($l->t('Sensitivity'));?>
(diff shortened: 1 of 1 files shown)


matiasdelellis:
Hi @Derkades

This is the reason why I never finish updating the translations of your changes.. Sorry about it... :disappointed: 

Do you want to comment on these new texts?

Derkades:
It looks fine, let me know when you push to transifex so I can translate them

matiasdelellis:
Done,
In the next release (I hope soon) we will have better translations. :smile

PR #445, opened 2021-03-19
l10n: Add a conjunction

Valdnet:
Parts of the sentence connected by the conjunction "and", as they are equivalent sentences.


matiasdelellis:
Yes.. Much better,
If you have more suggestions they are always welcome.  

Thanks :smile:

PR #508, opened 2021-08-29
Added percentage into Progress command

gradrix:


matiasdelellis:
Hi @gradrix 
It's a good contribution ... Thanks.. 😄 
It looks good, but I will ask you that try to respect the styles. You are using spaces instead of tabs..

Also.. You can include yourself to the coyright in these file... 😉

PR #511, opened 2021-09-06
Migrate hooks to OCP event listeners

nkming2:
This PR migrates the file hooks to OCP event listeners instead. No change to the code handling the events.

There are two reasons behind this PR. First, hooks are now deprecated according to https://docs.nextcloud.com/server/latest/developer_manual/basics/events.html. Second, registering hooks in boot caused $UserId to be null when injected to the Controllers in some situations. To reproduce it, simply add @NoCSRFRequired to PersonController::index and execute a curl cmd. The $UserId will be null when injected in the constructor. My understanding to the issue is that, FileHooks depends on $UserId, but at that moment, the user session has not yet been set by Nextcloud if authenticating by username and password. After that, the DI container is probably caching the resolved value and subsequent constructs of the Controllers would also receive null. This may sound a non-issue right now since the API is not working anymore, but it'll also affect the future OCS API when implemented.

I've also noticed that currently FileHooks is triggered when putting files to trash but not triggered when restoring from trash. I'll leave this to another PR later as it kind of depends on this PR.

### DIFF
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -69,4 +69,7 @@ class Application extends App implements IBootstrap {
 		$context->registerEventListener(NodeDeletedEvent::class, PostDeleteListener::class);
 	}
 
+	public function boot(IBootContext $context): void {
+	}
+
 }
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -29,6 +29,7 @@ namespace OCA\FaceRecognition\AppInfo;
 
 use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootstrap;
+use OCP\AppFramework\Bootstrap\IBootContext;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
 
 use OCP\EventDispatcher\IEventDispatcher;
--- a/lib/AppInfo/Application.php
+++ b/lib/AppInfo/Application.php
@@ -30,17 +30,17 @@ namespace OCA\FaceRecognition\AppInfo;
 use OCP\AppFramework\App;
 use OCP\AppFramework\Bootstrap\IBootstrap;
 use OCP\AppFramework\Bootstrap\IRegistrationContext;
-use OCP\AppFramework\Bootstrap\IBootContext;
-use OCP\AppFramework\IAppContainer;
 
 use OCP\EventDispatcher\IEventDispatcher;
 
 use OCA\Files\Event\LoadSidebar;
+use OCP\Files\Events\Node\NodeDeletedEvent;
+use OCP\Files\Events\Node\NodeWrittenEvent;
 use OCP\User\Events\UserDeletedEvent;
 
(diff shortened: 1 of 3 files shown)


matiasdelellis:
Hi @nkming2 

Thank you very much for this PR. It looks good, and now I'm going to test it. 😄 
You can include the copyright in these files.. 😉 

> I've also noticed that currently FileHooks is triggered when putting files to trash but not triggered when restoring from trash. I'll leave this to another PR later as it kind of depends on this PR.

You are right, it is something that I have pending, and I would be happy to accept that PR.. 😄

matiasdelellis:
Hi @nkming2 


```
{
   "reqId":"YTqTnQvTPg5gjw7bXGLvDgAAAE0",
   "level":3,
   "time":"2021-09-09T23:07:09+00:00",
   "remoteAddr":"190.48.181.47",
   "user":"--",
   "app":"PHP",
   "method":"GET",
   "url":"/apps/dashboard/",
   "message":"Class OCA\\FaceRecognition\\AppInfo\\Application contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (OCP\\AppFramework\\Bootstrap\\IBootstrap::boot) at /var/www/html/nextcloud/apps/facerecognition/lib/AppInfo/Application.php#48",
   "userAgent":"Mozilla/5.0 (X11; Fedora; Linux x86_64; rv:91.0) Gecko/20100101 Firefox/91.0",
   "version":"21.0.4.1"
}{
   "reqId":"y0OhWAQy362n2noLWB65",
   "level":3,
   "time":"2021-09-09T23:07:31+00:00",
   "remoteAddr":"",
   "user":"--",
   "app":"PHP",
   "method":"",
   "url":"--",
   "message":"Class OCA\\FaceRecognition\\AppInfo\\Application contains 1 abstract method and must therefore be declared abstract or implement the remaining methods (OCP\\AppFramework\\Bootstrap\\IBootstrap::boot) at /var/www/html/nextcloud/apps/facerecognition/lib/AppInfo/Application.php#48",
   "userAgent":"--",
   "version":"21.0.4.1"
}
```

lib/AppInfo/Application.php
```diff
@@ -64,10 +64,9 @@ public function register(IRegistrationContext $context): void {
 
 		$context->registerEventListener(LoadSidebar::class, LoadSidebarListener::class);
 		$context->registerEventListener(UserDeletedEvent::class, UserDeletedListener::class);
-	}
 
-	public function boot(IBootContext $context): void {
```
matiasdelellis, line 69: Well, basically requires restoring this method. 🤔

matiasdelellis: Ok... After restoring this method, I was able to test it, and everything seems to work properly. 😄 

Please, add that.. 😉

nkming2: Oops. I left part of the old code to compare the difference so I didn't catch that, my bad. :persevere:

matiasdelellis:
Great.. Thanks again @nkming2  😄 🎉

PR #512, opened 2021-09-06
OCS API to get all persons

nkming2:
This PR implements the /persons endpoint from the old API as OCS API.

The returned data is identical to the old API in https://github.com/matiasdelellis/facerecognition/wiki/RESTful---Endpoint---API#get-all-persons-with-faces-and-file-asociated. The new endpoint would be /ocs/v2.php/apps/facerecognition/api/v1/persons

### DIFF
--- a/lib/Controller/PersonApiController.php
+++ b/lib/Controller/PersonApiController.php
@@ -1,0 +1,117 @@
+<?php
+/**
+ * @copyright Copyright (c) 2018-2020 Matias De lellis <mati86dl@gmail.com>
+ *
+ * @author Matias De lellis <mati86dl@gmail.com>
+ *
+ * @license GNU AGPL version 3 or any later version
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as
+ * published by the Free Software Foundation, either version 3 of the
+ * License, or (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+namespace OCA\FaceRecognition\Controller;
+
+use OCA\FaceRecognition\Db\Face;
+use OCA\FaceRecognition\Db\FaceMapper;
+use OCA\FaceRecognition\Db\Image;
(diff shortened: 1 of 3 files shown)


nkming2:
Related to #511

matiasdelellis:
Hi  @nkming2 
Again, thank you very much for the contribution..

I understand that you are doing the API for an application. Being the first customer for this API, I think it is a good time to discuss this.

I am very concerned about the performance of the server.. 🤔

lib/Controller/PersonApiController.php
```diff
@@ -0,0 +1,117 @@
+<?php
+/**
+ * @copyright Copyright (c) 2018-2020 Matias De lellis <mati86dl@gmail.com>
+ *
+ * @author Matias De lellis <mati86dl@gmail.com>
```
matiasdelellis, line 5: You are the author.. 👏🏻 👏🏻 😉

lib/Controller/PersonApiController.php
```diff
@@ -0,0 +1,117 @@
...
+		$this->urlService      = $urlService;
+		$this->userId          = $UserId;
+	}
+
+	/**
+	 * Get all persons with faces and file asociated
+	 *
+	 * @NoAdminRequired
+	 */
+	public function getPersons() {
```
matiasdelellis, line 83: You are doing a simple query, which returns all the persons, all the faces and all the images.. 🙈 

This can be simple to consume in an application, but very very hard on the server if you have 50,000 photos, 😞

matiasdelellis: Can you do something similar to the main application?

1. Get the "persons"...
2. When necessary, the photos of a person.
3. When necessary,  the faces/persons in a photo.

matiasdelellis (changes requested):

lib/Controller/PersonApiController.php
nkming2: > You are doing a simple query, which returns all the persons, all the faces and all the images

I was following the old API hoping that it'll be accepted more easily :wink:. If it's ok to change it, I would also prefer splitting it into multiple APIs

> but very very hard on the server if you have 50,000 photos

Hard to say. Currently there's not much data per row, so I believe the performance impact shouldn't be that much. But nevertheless, I agree that this is not a good approach

nkming2:
Done. It's now one endpoint to list all persons, and another to list faces of a single person

matiasdelellis:
Well, It is an api that can work properly.. So, I will merge this PR... 

Of course, you are still the only potential client, and I am willing to continue discussing it.. 😉 

Thanks for all,
Regards,
Matias,

nkming2:
Thanks for merging it and also for this great app! :smile: Hopefully I could make more people learn about this great app after integrating the API :wink: :relaxed:. Thanks again :+1:

Issue #527, opened 2021-10-29
Discrepancy in allowed and displayed temporary file sizes.


guystreeter:
Hey, Thanks for reporting issues back to Nextcloud Face Recognition. Please, try to complete this report in detail so we can help you easier. :smile:

Make sure you read all the [documentation](https://github.com/matiasdelellis/facerecognition/wiki), and the [FAQ](https://github.com/matiasdelellis/facerecognition/wiki/FAQ), and that the issue has not been reported before. :wink:

### Expected behaviour

I should be able to set the file size slider to the maximum allowed by the model.

### Actual behaviour

The UI will tell me I can't set it higher than the value I'm trying to set it to.

### Steps to reproduce
1. Using model 4, try to set the slider to 1672x1254
2. The UI will tel you that you cannot set it higher than 1672x1254

### Server configuration
Latest nextcloud apache docker image, with self-built pdlib, and current git clone of facerecognition

### Additional information
The problem appears to be the rounding that happens to create the 4x3 values. The slider can be set to something that displays 1672x1254 but is fractionally higher, because the display rounds down.

I think a reasonable fix is to keep the displayed maximum value in the error message rounded down as it is now, but round up on the slider's displayed value.

guystreeter:
Another, possibly better, solution: when the user clicks the save button, convert the slider value to rounded-down width and height, then multiply those get a value to be set.

guystreeter:
A third option: set the max value of the slider to the maximum allowed by the model, since you don't let the user set it any higher than that anyway. I've never tried to write javascript, so I have no idea how to implement that.

matiasdelellis:
Thank again @guystreeter 🎉 😄 
 
I couldn't test it, but it was 100% logical.. 😬 
When you want, close this report yourself. 😉 

p.s: Quick answer... I hope to have free time soon to do a few things before NC23

PR #528, opened 2021-10-31
Set the Image Area slider max value to the maximum allowed by the model

guystreeter:
Saving a value higher than the model's maximum is disallowed
anyway, and this resolves the problem reported in #527 

I'm just learning PHP, so let me know if I can improve on this.


matiasdelellis:
Wow @guystreeter
I guess it's a good solution... Very simple, that solves something annoying. 😬 

Thank you very much for that.. 😄

PR #534, opened 2021-11-24
Setup memory

matiasdelellis:
I add an additional step to the configuration, but that avoids unnecessary confusion..

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -79,6 +79,7 @@ jobs:
       - name: Upload coverage to scrutinizer-ci
         working-directory: apps/${{ env.APP_NAME }}
         run: |
+          ../../occ face:setup --memory 1G
           ../../occ face:setup --model 1
           make test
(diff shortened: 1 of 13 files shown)


matiasdelellis:
Well.. in short.. An initial step is added.. 😅 

You must specify exactly how much memory you want to assign for image processing.

```
[matias@nube nextcloud]$ sudo -u apache php occ face:setup -M 3G
[sudo] password for matias: 
System memory: 7.2 GB (7726276608B)
Memory assigned to PHP: Unlimited

Minimum value to assign to image processing.: 682.7 MB (715827882.66667B)
Maximum value to assign to image processing.: 4.8 GB (5150851072B)

Maximum memory assigned for image processing: 3 GB (3221225472B)
```

**Note:**
* This command must be executed extensively by all users, even those who are currently using the application.
* Now the [memory_limit](https://www.php.net/manual/es/ini.core.php#ini.memory-limit) conf should be done only by the command line process. There is no need to touch more on the web server configuration.
* It is designed primarily to avoid user confusion. If you have any questions, you should check these commands!.

Issue #572, opened 2022-04-16
The external model should not require memory setup


guystreeter:
I am trying to use the external model so I can run nextcloud on a server with limited memory. The external model does not use much memory, but the app fails to set up due to the limited memory of the system.

matiasdelellis:
As comment in #573 the external model don't consumes memory, but the clustering is done locally and if you have many faces, it can still consume a lot of memory. Also the documentation should be generic, and the models must be configured in the same way.

Also as I said, I have to review the documentation a bit, and perhaps the interaction/configuration of the external model. That's why I'm going to leave open this issue, but memory control/settings is necessary. 🤔 

Thanks for all,

PR #573, opened 2022-04-16
Skip memory requirements tests when the external model is used.

guystreeter:
This is a suggested fix for #572 
It skips checking for memory setup when the model ID is 5.
Maybe there's a better way to do it? I'm open to suggestions.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -13,9 +13,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['7.3']
+        php-versions: ['7.4']
         databases: ['mysql']
-        server-versions: ['stable22']
+        server-versions: ['stable24']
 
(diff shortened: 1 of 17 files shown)


matiasdelellis:
Hi @guystreeter 
Thanks for the change, but I'm not going to merge it. The external model don't consumes memory, but the clustering is done locally and if you have many faces, it can still consume a lot of memory.

Also the documentation really should be generic. and the models must be configured in the same way. On the other hand, recognize that needs some clarification about the external model (ie. the service must be available to configure the size of the temp image), but removing checks is not the way to go.

guystreeter:
My experience is that the external model runs fine with 1GB system memory. The main reason I am using the external model is that cloud providers charge a lot more for a 2GB server.

matiasdelellis:
> My experience is that the external model runs fine with 1GB system memory. The main reason I am using the external model is that cloud providers charge a lot more for a 2GB server.

Yes.. But you have to solve it in another way. Maybe I have to review all the process, but you can't ignore this setting just because it's an external model. 😅 

We continue at #572

PR #621, opened 2022-12-26
Rename identifiers in OCS Api controller to match what Nextcloud expects

nkming2:
Right now calling any of the OCS APIs in v0.9.10 will throw a OCP\AppFramework\QueryException on the server, because the file, class and method are not named correctly. This PR fixed them such that the router can correct instantiate the class and invoke the correct method.


matiasdelellis:
Wow..
As always, thank you very much. 😄

PR #709, opened 2023-11-16
added options to command face:sync-albums

wronny:
Hello,
I like facerecognition and its sync-album command. Because I don't want/need to create albums for all persons I added an option to create only albums for selected persons. I also like to create albums with photos containing several persons therefore I added a mode option "Album-per-person" vs "Album-combined". Also added a list persons option to allow to list persons from command line.

Options added to command face:sync-albums:
    -l, --list_persons : List all person defined for a given user_id
    -p, --person_name=PERSON_NAME : Sync albums for a given user and person name(s) (separeted using comma). If not used, sync all persons defined by the user
    -m, --mode=MODE : Album creation mode. Use "Album-per-person" to create one album for each given person via person_name parameter. Use "Album-combined" to create one album for all person names given via person_name parameter. [default: "Album-per-user"]

I hope this is meaningful/helpful also for others.  Of course open for feedback / discussion. 

Cheers,
Marcel

### DIFF
--- a/lib/Command/SyncAlbumsCommand.php
+++ b/lib/Command/SyncAlbumsCommand.php
@@ -176,6 +176,10 @@ class SyncAlbumsCommand extends Command {
 				}
 				else if ($mode === "Album-combined") {
 					$personList = explode(",", $person_name); 
+					if ( count($personList) < 2 ){
+						$output->writeln("Note parameter mode <$mode> requires at least two persons");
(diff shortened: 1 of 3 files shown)


matiasdelellis:
Well, 
I was never able to test it properly, but it's a nice addition, and doesn't change the previous flow. 

Thanks again,

Issue #91, opened 2018-12-30
Support to give specific path to command


stalker314314:
Say, user wants to process just some subdirectory, it would be good if we can support
```
./occ face:background_job --path /files/username/pics/just_one_vacation
```
It will affect some tasks (where applicable), but some will be indeferent to this.

PR #739, opened 2024-04-08
Force analyze files

36grad:
Add an option (--force_analyze_files) that enablee the user to explicitly (re-)analyze specific files.
    
    - The  --force_analyze_files option allows the user to specify files that they want to analyze.
    - The  --force_analyze_files (or -f) option can be used multiple times to analyze more than one file.
    - The images must be given relative to the data/USER_ID/files folder and thus the --user_id option is required.
    - The  --force_analyze_files option does not work together with sync-mode or cluster-mode.
    - If no mode is specified by the user, defer-mode is enabled for a better user experience.
    - Files that are specified using the  --force_analyze_files option are put at the front of the analysis queue and thus are analyzed first.

### DIFF
--- a/lib/BackgroundJob/BackgroundService.php
+++ b/lib/BackgroundJob/BackgroundService.php
@@ -80,10 +80,11 @@ class BackgroundService {
 	 * @param IUser|null $user ID of user to execute background operations for
 	 * @param int|null $maxImageArea Max image area (in pixels^2) to be fed to neural network when doing face detection
 	 * @param string $runMode The command execution mode
+	 * @param array $forceAnalyzeFiles An array containing the names of files that shall be analyzed
 	 *
 	 * @return void
 	 */
-	public function execute(int $timeout, bool $verbose, IUser $user = null, int $maxImageArea = null, string $runMode) {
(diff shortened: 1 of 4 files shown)


matiasdelellis:
Hi @36grad 
Again, Thanks for this PR.

Can I ask you for a change in approach to the problem?. Instead of adding a force option, (which ultimately resets the file), wouldn't it be better to add an option to face:reset to reset files? 🥲

36grad:
Yep, that is a good suggestion - I had worked initially based on the PHP code from the appstore version and when I pulled the latest source from the repository I saw that I had to add a code line where I reset the file's status - so this would have been a good approach in the first place.
There's one thing, however, which I would be missing is having the option to explicitly analyze those files. If they're only reset there's still a huge number of files to be analyzed there's no way to tell when the reset files will be analyzed. Any ideas here?

matiasdelellis:
> If they're only reset there's still a huge number of files to be analyzed there's no way to tell when the reset files will be analyzed. Any ideas here?

MMm.. 🤔 
In any case you could add a [--path option](https://github.com/matiasdelellis/facerecognition/issues/91) and you become a hero!. :wink:
In the end you will reuse a lot of what you made, but IMHO it makes makes more sense to reset and scan explicitly instead of forcing something.

nicolas-r:
Hello

May I ask what is the status of this merge request ?

Issue #700, opened 2023-10-01
Imaginary service?


j3ffery17:
Hey, Thanks for reporting issues back to Nextcloud Face Recognition. Please, try to complete this report in detail so we can help you easier. :smile:

Make sure you read all the [documentation](https://github.com/matiasdelellis/facerecognition/wiki), and the [FAQ](https://github.com/matiasdelellis/facerecognition/wiki/FAQ), and that the issue has not been reported before. :wink:

### Expected behaviour

Trying to run initial background job 

     sudo -u www-data ./occ face:background_job -u user
 
### Actual behaviors

Getting the following:

1/8 - Executing task CheckRequirementsTask (Check all requirements)
        An Imaginary service (http://127.0.0.1:8080) was configured to manage temporary images, but it is inaccessible.Check out the service, or set the 'preview_imaginary_url' key appropriately.

### Steps to reproduce
1.
2.
3.

### Server configuration

* **Operating system:**

* **Pdlib version:**

* **How is DLib installed:** Make sure it is working correctly with this [tool](https://github.com/matiasdelellis/pdlib-min-test-suite)

* **How is PDlib installed:** Make sure it is working correctly with this [tool](https://github.com/matiasdelellis/pdlib-min-test-suite)

* **PHP version:**

* **Web server:**

* **Database:**

* **Nextcloud version:**

### Client configuration

* **Browser:**

* **Operating system:**

#### Background task log with debug.
<details>
<summary>sudo -u apache php occ -vvv face:background_job</summary>

```
Insert your background log here
```
</details>

#### Web server error log
<details>
<summary>Web server error log</summary>

```
Insert your webserver log here
```
</details>

#### Nextcloud log (data/nextcloud.log)
<details>
<summary>Nextcloud log</summary>

```
Insert your Nextcloud log here
```
</details>

#### Browser log
<details>
<summary>Browser log</summary>

```
Insert your browser log here, this could for example include:

a) The javascript console log
b) The network log
c) ...
```
</details>

alsoszaa:
I think I had this issue as well because I was running a local instance of Nextcloud. 
Go to /var/www/NextCloud directory and edit the config.php file to allow local instances

'allow_local_remote_servers' => true,

j3ffery17:
Rhabks. Truied that and still saw the error wo I commented out rhe preview_imaginary_url line in config.php and then it worked.

GuillaumePod:
[...]

PR #746, opened 2024-04-30
Add key support for Imaginary

fabalexsie:
This small PR fixes https://github.com/matiasdelellis/facerecognition/issues/700 where facerecognition is used with Imaginary and Imaginary is configured so that you must provide a key to use it (see [h2non/imaginary Authorization](https://github.com/h2non/imaginary?tab=readme-ov-file#authorization)).

- [X] test with Imaginary requiring a key
- [ ] test with Imaginary without requiring a key (nothing should change)

Unfortunately, I don't have a setup for the latter, so I can't test it myself.

Note: The authorization is implemented by adding the key as query parameter to the requests. As stated in the [imaginary readme](https://github.com/h2non/imaginary?tab=readme-ov-file#authorization), it could also be done via an HTTP header as an alternative.

### DIFF
--- a/lib/Helper/Imaginary.php
+++ b/lib/Helper/Imaginary.php
@@ -55,6 +55,19 @@ class Imaginary {
 		return rtrim($imaginaryUrl, '/');
 	}
 
+	public function hasKey(): ?string {
+		$imaginaryKey = $this->config->getSystemValueString('preview_imaginary_key', 'invalid');
+		return ($imaginaryKey !== 'invalid');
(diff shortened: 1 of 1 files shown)


matiasdelellis:
Wow @fabalexsie
I just never see that nextcloud recommend it, so I never took it into account. 😅 

p.s. I also have some reservations with the use of imaginary. Personally, the only reason for implementing its support was to be able to read HEIC images and today it is not even necessary.

PR #750, opened 2024-05-16
Add the sixth model to the application. Aka DlibTaguchiHog model. =) 

matiasdelellis:
In this model 6, we use the model newly trained by Tanguchi to obtain
the descriptor and perform the face recognition.

This model was trained from scratch to slightly improve the bias of the
original model on non-Caucasian/American people. It obtained a similar
result in the LFW tests, slightly lower, but within the acceptable
margins of error.

It may still have bias problems, they recognize that it fails with
African people and with childrens, but in principle it seems like an
improvement to the model used in 1, 2, 3 and 4. 😉

See: https://github.com/TaguchiModels/dlibModels

This model is added for testing, and I understand that it will be of
great help to those who have had problems with the original model.

### DIFF
--- a/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
+++ b/lib/Model/DlibCnnHogModel/DlibCnnHogModel.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * @copyright Copyright (c) 2020-2023, Matias De lellis <mati86dl@gmail.com>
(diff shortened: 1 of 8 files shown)


RobertaAndrea:
i get this error ERROR: Unexpected version found while deserializing dlib::add_layer. 
when using dlib_face_recognition_resnet_model_v1.dat  in code:         
// Load the pretrained face recognition model (make sure the model file is available)
        anet_type net;
        deserialize("dlib_face_recognition_resnet_model_v1.dat") >> net;

PR #761, opened 2024-07-04
English language improvements

mhameed:
Thank you for your time in creating and developing this app.
Please consider the language improvements suggested in the commit.
Thank you.


matiasdelellis:
▎ Thanks for taking the time.
▎
▎ I took the README changes in <commit>. I left out the ones on the interface strings: they are message ids used by the translations, so changing "Analyze" to "Analyse" or renaming "Minimum confidence" invalidates the work of the translators for a difference in style. Also watch out, the patch introduced a typo ("alalysis").

PR #778, opened 2024-10-01
Add support for NC30 and deprecate NC28

matiasdelellis:


vwbusguy:
For what it's worth, I'm about to create clusters with the cron on Nextcloud 30 on Mariadb 10.11.9 and PHP 8.3.12 (Fedora 40).  However, the dialog is broken for me (#779) so maybe I'm not triggering the method that is actually failing here.  At least it appears to be a consistent failure across the board.

Since I have a populated Nextcloud 30 environment, let me know if I can be of any help with testing and debugging.  This Nextcloud app has been very useful for me, and I'm happy to pitch in where I can be useful here.

vwbusguy:
So close!

PR #820, opened 2025-09-14
Feature/improve shared file handling

mate1213:
This PR solve the handling of shared, external image
Before this PR the shared images stored and processed multiple times.

This become a problem when you have large number of files (ex. 40 000+) and this has been share between 5-10 accounts. The processing time of these files could take a long time (like 6-8 months if there is no GPU acceleration).

For this I needed to change the Database Scheme. The new schema can be found in [ERD-Schema.png](https://github.com/matiasdelellis/facerecognition/pull/820/files#diff-ed363980bc18ef6d03c1386c4fa4401c000f4eb49b26a7363b7192fd90731bb1)

I have extended the Unit tests and made the Integration tests Updated to be able to run with the latest NC update (NC31)


mate1213:
With indepth testing with NC32 I've found multiple bugs Which I need to resolve first

mate1213:
I have made it stable in NC31 andalsoin NC32 now this should be ready for review

NL-TCH:
@matiasdelellis can you please review and merge?

matiasdelellis:
Hi, thank you so much for your contribution. It's finally implemented in master now! 😄 

Justification by Claude.. ?
▎ First of all, thank you for the amount of work here, and sorry it took me so long to get to it.
▎
▎ The problem it attacks is real, but it is already solved: a shared photo is no longer analyzed once per user. ImageProcessingTask looks for an already processed image with the same file id (ImageMapper::findProcessedDuplicate()) and copies its faces (FaceMapper::copyFaces()) instead of running the model again. It is covered by tests/Integration/ReuseAnalysisTest.php.
▎
▎ That leaves this PR with one row per user in facerecog_images as the thing it would still improve, and I can't justify a schema change of this size, in 105 files, for that. I'd rather not carry it. If there is something else in here that the reuse doesn't cover, tell me and I'll look at it as a separate PR.

PR #825, opened 2025-10-03
Add ImageMagick support

MAFLO321:
This adds support for image types like heic, tiff, and many more without the need for hosting imaginary application.

This is a basic implementation that lacks extended mimetype checking.

### DIFF
--- a/lib/Helper/TempImage.php
+++ b/lib/Helper/TempImage.php
@@ -139,7 +139,7 @@ class TempImage extends Image {
 
 			$this->ratio = 1 / $scaleFactor;
 		}
-		else {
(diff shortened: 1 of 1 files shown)


matiasdelellis:
Done...  thanks.. 😄 

▎ Thanks for this, and sorry for the delay.
▎
▎ Support for these formats landed in the meantime, in b2b54da: ImageUtil::loadInto() tries GD, falls back to Imagick, and finally to Imaginary, which is the only one that reads HEIC or AVIF when Imagick was built without the delegates. The main difference with this patch is that the downscale happens during the decode, so the bitmap of a HEIC from a phone is never held whole in memory, which is where these formats hurt.

PR #838, opened 2026-04-15
Nc33 upgrade

nerdling:
Building on #830 this adds a commit that replaces QueryBuilder::execute()

PR #840, opened 2026-06-04
Support Nextcloud 33 (PHP 8.2–8.5)

miguelkmarques:
## Summary

  Brings the app to **Nextcloud 33** (PHP 8.2–8.5). Rebased onto EngTurtle's master,
  so it carries forward the existing HEIC and NC32 work; the three commits on top
  replace the OCP APIs NC33 removed and clear PHP 8.4 deprecations so the app enables
  cleanly on a live NC33 instance.

  ## Base

  Rebased onto `EngTurtle/facerecognition@master` (tip `66b860d`), which already
  includes HEIC thumbnail generation (`fb4a87a`, PR #2) and the NC32 line with its
  image-processing optimizations (PR #1). Those are inherited, not re-introduced here.

  ## Relationship to #838

  #838 also targets NC33 but is a **partial** fix — it only swaps `execute()` in the
  three Db mappers. Both branches were cut from the same base (`7752d69`), so the gap
  is exact:

  | Change | #838 | This PR |
  |---|:--:|:--:|
  | `execute()` in Db mappers (3 files) | ✅ | ✅ |
  | `execute()` in Migration files (4 files, ~12 calls) | ❌ | ✅ |
  | `info.xml` NC `max-version` → 33 | ❌ | ✅ |
  | `info.xml` PHP `max-version` → 8.5 | ❌ | ✅ |
  | `OC_App::getAppPath()` removal fix | ❌ | ✅ |
  | PHP 8.4 nullable-param deprecations | ❌ | ✅ |
  | `PersonController::find()` null guard | ❌ | ✅ |
  | `composer.json` / CI / CHANGELOG | ❌ | ✅ |

  Because #838 leaves `execute()` in the migrations and never bumps `info.xml`, it
  still fatals on `app:enable` (`Call to undefined method ...QueryBuilder::execute()`)
  and can't pass the NC33 version gate without `--force`.

  ## Changes (3 commits on top of master)

  ### Version / metadata
  - `appinfo/info.xml`: NC `max-version` 32 → 33, PHP `max-version` 8.4 → 8.5, `<version>` bump.
  - `composer.json`: `nextcloud/ocp` `dev-stable32` → `dev-stable33`.
  - CI: test against `stable33` / PHP `8.2–8.5`. CHANGELOG entry.

  ### Removed-API fixes (NC33)
  - `IQueryBuilder::execute()` removed → `executeQuery()` (SELECT) / `executeStatement()`
    (write) across `lib/Db/*Mapper.php` **and** `lib/Migration/Version*.php`.
  - `OC_App::getAppPath()` removed → inject `OCP\App\IAppManager`, use
    `$appManager->getAppPath('facerecognition')` (`FaceController`).

  ### PHP 8.4 deprecations
  - Explicit nullable types (`?IUser`, `?int`, `?string`) on null-defaulted params
    (`BackgroundService::execute`, `MigrateCommand::getEligiblesUserId`, `FileService::getLocalFile`).
  - Guard `PersonController::find()` against `current()` returning `false` (person with no faces).

  ## Testing
  - `make build-vue javascript-deps js-templates` builds assets.
  - `occ app:enable facerecognition` succeeds on NC33 with no `--force`.
  - End-to-end against an external model (model 5): sync → analyze → cluster run clean,
    faces detected and grouped.
  
  ## Notes
  - No DB schema changes — migrations only had `execute()` calls swapped.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         php-versions: ['8.3']
         databases: ['mysql']
-        server-versions: ['stable31']
+        server-versions: ['stable32']
 
     name: php${{ matrix.php-versions }}-${{ matrix.databases }}-${{ matrix.server-versions }}
 
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -15,7 +15,7 @@ jobs:
       matrix:
         php-versions: ['8.3']
         databases: ['mysql']
-        server-versions: ['stable32']
+        server-versions: ['stable33']
 
     name: php${{ matrix.php-versions }}-${{ matrix.databases }}-${{ matrix.server-versions }}
 
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -19,7 +19,7 @@ jobs:
       matrix:
         php-versions: ['8.1', '8.2', '8.3', '8.4']
         databases: ['sqlite', 'mysql', 'pgsql']
-        server-versions: ['stable31']
+        server-versions: ['stable32']
 
     name: php${{ matrix.php-versions }}-${{ matrix.databases }}-${{ matrix.server-versions }}
 
--- a/.github/workflows/phpunit.yml
+++ b/.github/workflows/phpunit.yml
@@ -17,9 +17,9 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-versions: ['8.1', '8.2', '8.3', '8.4']
+        php-versions: ['8.2', '8.3', '8.4', '8.5']
         databases: ['sqlite', 'mysql', 'pgsql']
-        server-versions: ['stable32']
+        server-versions: ['stable33']
 
     name: php${{ matrix.php-versions }}-${{ matrix.databases }}-${{ matrix.server-versions }}
 
--- a/.github/workflows/static-analysis.yml
+++ b/.github/workflows/static-analysis.yml
@@ -11,7 +11,7 @@ jobs:
       runs-on: ubuntu-latest
       strategy:
           matrix:
-              ocp-version: ['dev-stable31']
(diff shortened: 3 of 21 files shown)


EngTurtle:
Hey, this is a duplicate of #838, also you probably want to double check claude's work, the PR mentions my `fb4a87a` commit but that change is nowhere in this PR.

miguelkmarques:
> Hey, this is a duplicate of #838, also you probably want to double check claude's work, the PR mentions my `fb4a87a` commit but that change is nowhere in this PR.

Hey, thanks for taking a look. You're right about fb4a87a.

  That mention was my mistake. My branch was based on an older commit (7752d69),
  before fb4a87a in the history, so the HEIC change wasn't included even though it
  was already in your repo. I've rebased onto your current master now, so fb4a87a is
  in and the description is fixed.

  On #838 being a duplicate: I did look at it, and it's actually a partial fix. It
  only replaces execute() in the three Db mappers, but it leaves the same execute()
  calls in the four Migration files, so app:enable still fatals with the undefined
  QueryBuilder::execute() error on NC33. It also doesn't bump info.xml, so the
  version gate still blocks NC33 without --force.
  
  This PR covers all of that: execute() in the migrations too, the info.xml/PHP
  version bumps, the removed OC_App::getAppPath(), the PHP 8.4 nullable deprecations,
  and a null-person-face guard. Happy to consolidate if #838 gets brought up to the
  same coverage, but as it stands it won't enable on NC33.

matiasdelellis:
Hi everyone,
I'm going to merge this, just to move forward quickly.

I have very little free time, and even less to dedicate to my projects, but I promise to make progress soon. 😉 

Thank you for your patience and all your contributions!. 😄

vwbusguy:
Thanks for this - I don't see the new release in GitHub or Nextcloud Apps yet, and the current "stable" version is now v34, though I'm hoping supporting v34 from this point should be easier than the effort to get it up to v33.

PR #841, opened 2026-06-08
Feature/manually mark faces

### LINKED ISSUE #168: Manually Tagging Photos
Is there a current feature where you can tag photos manually?

Ex: Remove or add a person to an individual picture.

If i goto a picture's persons tab in files, the only option I see is to rename.

HerbertHorst:
Add the option to tag a face manual using the memories gui. Fixes #168 

You can check if you want to use the manual tag for clustering. 
If checked yes it tries to get a face in your marked part and then checks if this is similar to an already found cluster. 
It only checks with unnamed clusters or clusters with the same name that you provided so it can not get mixed up with another cluster from another person that it thinks looks similar. If no face is found for clustering (f.e. like in my provided picture with my cat) it silently ignors it for clustering) 

I also upped the max version thats supported so I could test on my nextcloud to 33

This is the backend to this pr in memories https://github.com/pulsejet/memories/pull/1678

grafik
grafik


matiasdelellis:
Hi @HerbertHorst 
I have no major problem accepting the manually mark faces. But these should not be incorporated into any type of analysis. Most of facial recognition models rely heavily on discovery boxes, since they were trained on a particular box model.

HerbertHorst:
Hi @matiasdelellis 
I know very few about how this works and tried to understand this with a bit of google and claude code (almost everything in this PR is from claude Opus) so if I misssed something im happy to learn and remove that feature again. 
**But** 
I tried to work around that problem by adding a 40% margin to the user crop instead of just passing it to the CNN so it should have enough awairness to create its own discovery box. Only if the CNN finds a face, a descriptor will be calculated. All the feature does is passing the user cropped part (plus the 40% margin) instead of the entire image to the CNN and disables the minimum size a face needs to have for manual created boxes. The bounding boxes should therefore still be determined by the CNN. 
```text
┌────────────────────────────┐
│         +40% Margin        │
│    ┌───────────────┐       │
│    │ User-Rectangle│       │
│    │ (x,y,w,h) Face│       │
│    └───────────────┘       │
│                  Other-Face│
└────────────────────────────┘
```
**But** I found a Bug in the logic while trying to understand all that, when the CNN finds a face in the +40% margin and decides to use that instead of the middle of the part that the user croped it will create a descriptor for that and still show the User created bounding box instead of the new CNN counding box that is actually used. 
I changed that so that it will save the CNN provided bounding box instead of the user provided. 

If you still see problems in that logic maybe we can find a solution or I will remove it.

HerbertHorst:
Fixed for the new db changes. 

This is how the output looks like when executing "background_job" and manuall tagges where added with the checkbox that I want them processed. For testing I used faces that where to small before and some 'no faces' to check that when no descriptor is found nothing breaks:
9/9 - Executing task ManualFaceDescriptorTask (Compute descriptors for manually added faces flagged for clustering)
        Computing descriptors for 5 manual face(s) of user HerbertH
        Manual face 13155: descriptor computed, it will be used for clustering
        Manual face 13156: no face detected in the marked region, excluding it from clustering
        Manual face 13159: no face detected in the marked region, excluding it from clustering
        Manual face 13160: descriptor computed, it will be used for clustering
        Manual face 13141: no face detected in the marked region, excluding it from clustering

HerbertHorst:
@matiasdelellis 
Sorry for so long nothing done from my side and also my rushed first message in ~~probably~~ bad english. 
I would like to make sure I actually addressed your concern from June, because everything else in this PR depends on your answer. Thanks in advance!

Your point was that manually marked faces should not be used for analysis, since the models rely heavily on their own discovery boxes. I agree, and I tried to build the feature with this in mind so a user-drawn box never reaches the model as a
discovery box:

- A manual face is only ever considered when the user explicitly ticks "use  for clustering". Without the tick it is stored with `is_groupable = false` and no descriptor, so it is pinned to its person and nothing more — the clustering never sees it.
- When use for analysis is ticked, `ManualFaceDescriptorTask` crops the marked region plus a 40% margin out of the original image and runs the model's own detector on that crop. A descriptor is computed only if the CNN finds a face by itself.
- The box the CNN returns then replaces the user rectangle in the database (scaled back up and shifted by the crop offset into original-image pixels). So the stored box and the descriptor always describe the same face, and the box always comes from the model, never from the user.
- The one threshold I work around is the minimum face size (`minImageSide = 1`) for the crop. That is the entire point of cropping: a face too small to be detected in the downscaled full image is large enough at crop resolution. If you would rather keep the threshold, I can drop this.

On robustness:
Nothing throws. Every failure path — source file gone, crop fails, no face detected, empty descriptor, any unexpected exception — is caught, logs a single line, sets `is_groupable = false` and moves on. No fallback descriptor is ever fabricated and a single bad region can never abort the background job. The face simply stays pinned to the person the user named.

The output in my comment above is exactly that case:
5 manual faces, 2 descriptors computed, 3 regions with no detectable face excluded, job finished as expected. 

Thanks for taking a look.

Unrelated to the code: the red Scrutinizer check is an environment issue as far as I can see, not this PR. Since the build image moved from bionic to jammy, PHP 7.3.15 (the default when no version is pinned) can no longer compile against OpenSSL 3.0,
so the build aborts before any test runs. Pinning a PHP version in `.scrutinizer.yml` maybe fixes that.

